Issue #34, opened 2020-04-28
ValueError: Each lower bound must be strictly less than each upper bound.


frigusgulo:
There seems to be an issue with the variogram fitting procedure. In "scipy/optimize/minpack.py", the call to the "least_squares"  is returning the error "ValueError: Each lower bound must be strictly less than each upper bound". This inputs are randomly selected points from a digital elevation model.

mmaelicke:
Hey Frankie,

Thanks for submitting this issue.
I will need some more information on this. Do you have a minimal example to reproduce the error? 
Can you also please submit a copy or a screenshot of the traceback?

What I can say is that the fitting procedure of the `Variogram` class will set initial guesses for the variogram parameters from the samples. But I cannot imagine how the lower bound can be set larger than the upper bound, unless all observations used (inside maxlag setting) are `NaN`.

You can try as a first guess into the blue, try to switch fitting method to `fit_method='lm'`, which will use Levenberg-Marquardt least squares.

Best,
Mirko

frigusgulo:
![Screenshot from 2020-04-28 14-58-21](https://user-images.githubusercontent.com/43794768/80537533-779f8f00-8961-11ea-95af-078e19fbfbfc.png)

frigusgulo:
This is the screenshot. I filtered for NaN values for the response value inputs, but it could be likely that the response vector is all zeros. I will switch to Levenberg-Marquardt and see how it fares.
Thanks for this excellent library!

mmaelicke:
Hey,
Do you have a code example for reproducing the error?

What I can see from your Traceback is:  you are passing in the `maxlag=int(m/2)`. I can't see what ' m'  is, I just can note that maxlag is the maximum *distance* at which point pairs are considered for vairogram estimation. So for the case that `m > 2`, which would be true if m is the number of samples or something like that, maxlag would be larger than 1 and following the [documentation of the `Variogram.__init__`(https://mmaelicke.github.io/scikit-gstat/reference/variogram.html#skgstat.Variogram.__init__) be interpreted as an **absolute** maximum distance (using the units of your coordinates). 

If that is the case, no point pairs are found within maxlag. The default fitting procedure will set initial guesses for the parameters in [__get_fit_bounds](https://github.com/mmaelicke/scikit-gstat/blob/566b1bbfa635484ce6f58ee1100c76f214877e64/skgstat/Variogram.py#L1166).

this looks like:
```python
[...]

PR #36, opened 2020-05-01
add Warning

mmaelicke:
Following #34 it can be helpful if a `Warning` is raised if all `values` passed to a `Variogram` are of same value. This PR includes this Warning for `Variogram` and `DirectionalVariogram`. Not sure how to implement on `SpaceTimeVariogram` yet. 

unittests missing. Will add them before merged.


mmaelicke:
`Variogram` will now raise a `Warning` if all input values are the same.

PR #40, opened 2020-09-15
Make variograms picklable

redhog:
Python pickle.dump() and joblib.dump() can not store objects with references to lambda functions. This is a work around to be able to persist variograms once created so that they can later be restored. This is important as calculating a variogram can be a pretty expensive operation, and it is therefore useful to do this once, and then later on be able to run kriging for various datasets using a variogram loaded from a file or a database.

### DIFF
--- a/skgstat/Variogram.py
+++ b/skgstat/Variogram.py
@@ -594,6 +594,12 @@ class Variogram(object):
     def dist_function(self):
         return self._dist_func
 
+    def _dist_func_wrapper(self, x):
+        if callable(self._dist_func):
+            return self._dist_func(x)
+        else:
+            return pdist(X=x, metric=self._dist_func)
+    
     @dist_function.setter
     def dist_function(self, func):
         self.set_dist_function(func=func)
@@ -626,7 +632,7 @@ class Variogram(object):
                 raise NotImplementedError
             else:
                 # if not ranks, it has to be a scipy metric
-                self._dist_func = lambda x: pdist(X=x, metric=func)
+                self._dist_func = func
 
         elif callable(func):
             self._dist_func = func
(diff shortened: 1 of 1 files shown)


mmaelicke:
Hey there,
Thank you very much for contributing!
Unfortunately, this change is breaking some tests. I don't think it's something serious, just a still existing reference to the old `dist_func` function, which now does not take the `metric` parameter anymore.

Here is the Travis output: 
https://travis-ci.org/github/mmaelicke/scikit-gstat/jobs/727343166

I will fix it myself and then merge the PR. This could take a couple of days, as I am buried in work right now. If you really really need the PR merged earlier, or want to check out the tests yourself, you can just drop another comment.

And thanks again!

redhog:
Hi!

Cool! I should have run the tests, not just tested things manually :S

I took a quick look, but it wasn't immediately obvious to me, so I'll just wait for you. I have a bunch of other changes I'll queue up for you too - they're definitely bound to break more stuff, so they'll require a conversation with you anyway...

mmaelicke:
Ok, I read your answer right after commenting on #44 :) 
Then first things first. I will look into the error of the unittests here first, as the pickling should not change functionality. Then we can look into the other PRs. Looking forward to work on the directional variogram stuff, as I finally did not use it in my PhD and it never got tested *in production*.

redhog:
Just out of curiosity, may I ask what your PhD thesis was about / what you used the variograms & kriging for?

mmaelicke:
> Just out of curiosity, may I ask what your PhD thesis was about / what you used the variograms & kriging for?

https://hess.copernicus.org/articles/24/2633/2020/

Well, after 3 years of rejected manuscripts and reviewing I didn't call them variograms anymore, and voila, publishing was not a problem anymore. I didn't do any kriging, that's why the variogram class has so much more functionality than the kriging class...

mmaelicke:
I opened a new PR with your changes and my adaptions: #45 

I don't know if I could have pushed directly into your PR, therefore I opened a new one and will close this one. Should be working now. If CI doesn't fail, I will merge #45 in a minute.

PR #43, opened 2020-09-17
Speed up the direction masking by > 2 orders of magnitude

redhog:
Tested using the dataset in `docs/tutorials/02_variogram_models.ipynb` but with the `DirectionalVariogram` both for correctness and for timing tests.

Time for 10 runs goes from 0:00:47.058498 to 0:00:00.590707 on my machine. This is comparable to a non-directional variogram for the same data.

**Note: This breaks quite a lot. In particular, only the compass and triangle directional model ares working right now.**

What this does: Instead of reprojecting all points once for every other point (a slow operation run many times), constructing a Shapely polygon and doing point-in-polygon tests (a very slow operation run many times), the difference vectors between point pairs are calculated in polar form (angle and distance) and used for filtering.

**Since this is a rather large functionality change, I'm posting this here mostly as a proposal for further discussion on optimizations.**

I have additionally included a plotting function, `pair_field(),` that plots a line segment for each point pair included in the directional mask for debugging purposes.


mmaelicke:
I will have a look as soon as your other PRs are merged. I am really curious and it sounds quite promising. I never really liked the avenue I took with shapely. I hope to find some time to look at it in detail by early next week.

I will also put this PR to the Version 0.3 milestone, as that was something I wanted to publish in the near future and as we are still < 1, we could introduce breaking changes with a minor version.

skgstat/DirectionalVariogram.py
```diff
@@ -452,20 +493,21 @@ def local_reference_system(self, poi):
...
 
         """
-        # define a point-wise transform function
-        def _transform(p1, p2, a):
-            p = p1 - p2
-            x = p[0] * np.cos(a) - p[1] * np.sin(a)
-            y = p[0] * np.sin(a) + p[1] * np.cos(a)
-            return np.array([x, y])
+
+        assert False
```
mmaelicke, line 497: Is this a mistake, or is there a reason to force the updated function to fail?

skgstat/DirectionalVariogram.py
```diff
@@ -31,7 +31,7 @@ def __init__(self,
                  normalize=False,
                  fit_method='trf',
                  fit_sigma=None,
-                 directional_model='triangle',
+                 directional_model='compass',
```
mmaelicke, line 34: Correct me if I am wrong, but at this state, there is only one directional model, right?

skgstat/DirectionalVariogram.py
```diff
@@ -550,6 +569,8 @@ def search_area(self, poi=0, ax=None):
         plot
 
         """
+        assert False
+
```
mmaelicke, line 573: Same here, is the `search_area` function meant to be removed?

skgstat/DirectionalVariogram.py
```diff
@@ -621,30 +660,10 @@ def _triangle(self, local_ref):
...
-            (xmax, -y),
-            (h, -y)
-        ])
-
-        return poly
-
-    def _circle(self, local_ref):
+
+        raise NotImplementedError
+
```
mmaelicke, line 665: On the one hand, I like the idea of getting rid of the shapely dependency, on the other hand, we are removing the circle, triangle and custom shaped search areas. At least the circle and triangle area shapes are published methods and I am wondering if we should really drop their support for the sake of calculation speed.
What do you think of keeping the old methods and add your code as a new *model*, called 'polar' or something?
It would be the default method and shapely could be an optional dependency. So if `directional_model` is not `'polar'`(or whatever it's called then), the class can check if shapely is installed and raise an error if not.

This way, `DirectionalVariogram` would be significantly faster, but the user can decide to be more flexible at the cost of speed, id needed.

mmaelicke (commented):
@redhog I added some thoughts. Would be great if you could clarify/comment. Thanks!

skgstat/DirectionalVariogram.py
redhog: Yep, this is just to make it work by default. I was planning for this to set back to "triangle" as soon as we have that implemented...

skgstat/DirectionalVariogram.py
redhog: So... that function isn't used anymore, and this was just a quick way to make sure it wasn't. The whole function should be removed...

skgstat/DirectionalVariogram.py
redhog: Yep... Or, it's used to plot the search area, so I'm unsure if we maybe should replace it with something else? I added a function to plot point pairs, maybe that's good enough for debugging?

skgstat/DirectionalVariogram.py
redhog: I think we could implement these two using the same methodology as the compass one - the math just gets a bit trickier...

For example, to constrain the angles to a rectangle around the azimith of badwidth width:

    lengh <= bandwidth / sin(angle - azimuth)

to create the triangle directional model, we'd need to combine this with something similar to what the compass one does,:

    (|angle - azimuth| <= tolerance) & (lengh <= bandwidth / sin(angle - azimuth))

redhog: I just wanted to implement the simple one first and see what you thought about it before I spent the time to implement all of them.

Oh, and I don't know how complicated the math gets for the circle on. Hm...

redhog: Just pushed a change to implement the triangle model and set that as default again. Wasn't too hard :)

redhog:
Added some comments, then implemented the triangle model using my new method (it's roughly as fast as the compass one)...

mmaelicke:
I see no reason why we shouldn't implement the new functionality. Maybe we push this PR into a new branch, until the at least the triangle is re-implemented? Then, I guess docs and unittests will need updates as well.
I am a bit busy today with my kids, but will come back to this PR this evening or tomorrow.

mmaelicke:
Here we go... didn't see your latest commit :)

redhog:
Oh, and please double check my math :P I've used my pair_field() function to make sure it looks realistic (comparing to the same plot on the old implementation), and I've done a few sketches on paper to make sure it's not totally crazy, but a second pair of eyes is always good :)

mmaelicke:
> Oh, and please double check my math :P I've used my pair_field() function to make sure it looks realistic (comparing to the same plot on the old implementation), and I've done a few sketches on paper to make sure it's not totally crazy, but a second pair of eyes is always good :)

I will do that. Maybe it would be good to have a test-dataset which could also be used in the unittest.

redhog:
Indeed it would. However, a small word of warning: There seem to be a small amount of rounding differences between the implementation, so there isn't a point for point match, as some points on the edge of the search field are handled differently :(

redhog:
I'm using my own dataset for testing, which I can't share, but you have some data in docs/data already - I guess we could just use that?

mmaelicke:
> I'm using my own dataset for testing, which I can't share, but you have some data in docs/data already - I guess we could just use that?

Ok, either we use that, or I will try generate a dataset later this evening and build an anisotropy into the field. That should make it easier to check the results. It could be used for the unittests as well and check the final variogram up to a specific precision, to overcome rounding errors.

redhog:
That sounds good! You could include some plots of that in the docs too to show of the directional kriging...

mmaelicke:
I opened #50 with all your changes included. I will add more (like fixing tests and adding docs). 
Should I make you a collaborator, then you can push to the new `directions` branch directly, without having to wait for me all the time?

redhog:
Cool! Yes, make me a collaborator and I'll poke at the unit tests etc.

Three of the current tests break on this branch - not sure why yet.

mmaelicke:
> Three of the current tests break on this branch - not sure why yet.

Don't waste too much time on them. It's absolutely possible, that they check against a wrong result. I programmed them quite quick and dirty at the time. I think a good pathway would be to check if the intermediate results are correct. If the mask array is correct for a small amount of points and if the number of point pairs in a specific lag class after masking is correct, I guess the unittests are fine.

PR #50, opened 2020-09-23
Direction

mmaelicke:
This is a copy of #43 to resolve merge conflicts, add test and verify that the calculations are correct.

I added a new field in the docs (`docs/data/aniso_x2.txt`) with an anisotropy showing more than twice correlation lengths in south-north that east-west direction. A sample from that field is added to the test folder, to build new tests (`skgstat/test/aniso_sample.txt`).
The whole stuff is in the new `directions` branch.

I checked the math and couldn't find anything obvious, nevertheless, with the given example and default parameters, I run into a divide by zero warning, that should be resolved.

```python
import numpy as np
import skgstat as skg

data = np.loadtxt('aniso_sample.txt')
north = skg.DirectionalVariogram(data[:,:2], data[:,2], normalize=False, azimuth=90, tolerance=45)
east = skg.DirectionalVariogram(data[:,:2], data[:,2], normalize=False, azimuth=0, tolerance=45)
```

@redhog, can you check where this warning is coming from? do you get it as well?

short checklist of things to do:

- [x] check the warning
- [x] add unittests for the different calculation steps
- [x] update docs
- [x] release as 0.3 version


redhog:
```
scikit-gstat/skgstat/DirectionalVariogram.py:311: RuntimeWarning: invalid value encountered in true_divide
  self._angles = np.arccos(scipy.spatial.distance.pdist(np.array([np.dot(_x, [1, 0])]).T, np.subtract) / self._euclidean_dist)
scikit-gstat/skgstat/DirectionalVariogram.py:582: RuntimeWarning: divide by zero encountered in true_divide
  return (absdiff <= np.radians(self.tolerance)) & (dists <= self.bandwidth / np.sin(np.abs(angles - np.radians(self.azimuth))))
```

This sounds like point pairs are formed between a point and itself, or another point at the exact same coordinate (_euclidean_dist == 0 and therefor their x components being the same => cos(x1 - x2) = 0).

mmaelicke:
OK, one approach would obviously be to skip calculations for the point to itself. Or, as two observations at the exact same positions are problematic anyway, include a check on instantiation if there are duplicates. Then, the above warning can just be catched and ignored.
`pdist` does not include the diagonal of the distance matrix. Maybe I sampled a point twice.... Would be bad luck. I will check and commit a new sample if true.

mmaelicke:
There was in fact one duplicated point. I draw a new sample from the same random field. So the properties of the variogram should be very similar.

redhog:
* Got all the unit tests to pass
* Manually verified that the directional_mask for a given dataset and settings (two different azimuths, both triangle and compass methods) did not change at all (not a single point treated differently) between the old and the new version.

redhog:
I think I'll need some help with getting the coverage bot happy and the docs up to scratch...

mmaelicke:
Well, don't mind the `0.03%` coverage drop. For the docs, I think I can update them on Wed + Thu, myself. If this PR is blocking other work you can also merge it anyway and open an issue to remind me of updating the docs.

redhog:
Nah, it isn't blocking me :) I've been working on our in-house software using this branch for some time now, I can stay on the branch for as long as we need to do this properly. Just wanted to clear up what the next step was and who's gonna do it :)

skgstat/DirectionalVariogram.py
```diff
@@ -277,6 +281,39 @@ def __init__(self,
...
+        # Calculate the angle between the vector between the two
+        # points, and east (see comment about self.azimuth)
+        # Note that this is in radians, while azimuth is in degrees.
+        # (a - b).[1,0] = ||a - b|| * ||[1,0]|| * cos(v)
+        # cos(v) = (a - b).[1,0] / ||a - b||
+        # cos(v) = (a.[1,0] - b.[1,0]) / ||a - b||
+        pos_angles = np.arccos(scipy.spatial.distance.pdist(np.array([np.dot(_x, [1, 0])]).T, np.subtract) / self._euclidean_dist)
+        # cos(v) for [2,1] and [2, -1] is the same, but v is not (v vs -v), fix that.
+        ydiff = scipy.spatial.distance.pdist(np.array([np.dot(_x, [0, 1])]).T, np.subtract)
+        self._angles = np.where(ydiff >= 0, pos_angles, -pos_angles)
```
mmaelicke, line 314: Hey @redhog ,
I think I have some trouble to understand this part, maybe you can help me to sort things out. This function is calculating the angle between the vector of a point pair and the x-axis, right?

I ran a short test to calculate some angles. Take the points `A = (0,0)`, `B = (1,2)` and `C = (1,2)`. If I pass them to the variogram, the condensed angles matrix `DirectionalVariogram._angles` should contain the angles of `AB` and `AC` to the x-axis as the first two elements.
I ran this code to do that:
```python
import numpy as np
from skgstat import DirectionalVariogram 

a = np.array([[0,0], [1,2], [2,1]])

# trying to run fit( ) causes an error for only 3 points, so I overwrite the method
class V(DirectionalVariogram):
  def fit(*args, **kwargs):
    pass

DV = V(a, np.random.normal(0,1,size=3))
DV._calc_direction_mask_data()
np.degrees(DV._angles)[:2]
```
and I get this result:
```
array([-116.56505118, -153.43494882])
```

But the results should be `~63°` for the angle between the vectors `[1,0]` and `[1,2]` , which would be the first element and `~26°` for the second.

I am not sure, if I don't get the function right or if I don't get the math right or if there is an error in our function.
With pen and paper I calculated for the first point:
Let `a` be the angle needed and `v = AB = [1,2]` and `e = [1, 0]` :
`cos(a) = (v . e) / ( |v| * |e| )`  (where `v . e` would be the scalar product)
then:
`cos(a) = (1*1 + 2* 0) / (sqrt(5) * sqrt(1)) = 1 / sqrt(5)`
and 
`a = arccos(1 / sqrt(5)) = 1.107 rad    =   1.107 * 180 / pi    =   63.4...°`

Am I blind? Do you see the mistake?

redhog:
```
>>> np.degrees(DV._angles)[:2] + 180
array([63.43494882, 26.56505118])
```

`scipy.spatial.distance.pdist` assumes distances are commutative and takes some shortcuts based on that that basically swaps the two points of a point pair sometimes. That means that the angles in `angles` are only correct +- 180 degrees, but since our filtering is symmetrical, I thought using pdist this way would be ok.

So no, you're not blind :)

docs/userguide/variogram.rst
```diff
@@ -609,13 +609,66 @@ Therefore, another step has to be introduced before lag classes are formed.
...
+
+SciKit-GStat implements a slightly adaped version of the formula given in the 
+last section. It makes use of symmetric search areas (tolerance is applied clockwise 
+and counter clockwise) und therefore any calculated angle might be the result 
+of calculating the orientation of :math:`\overrightarrow{AB}` or 
+:math:`\overrightarrow{BA}`. Mathematically, these two vectors have two different 
+angles, but they are always both taken into account or omitted for a variagram 
+at the same time. Thus, it does not make a difference for variography. 
+However, it does make a difference when you try to use the orientation angles 
+directly as the containing matrix can contain the inverse angles.
```
redhog, line 671: Awesome to have that documented :)

redhog:
:+1:

Issue #51, opened 2020-09-23
This can never have worked? Triangle calculation...


redhog:
While debugging why I get slightly different results with the new directional models, I came across this:

![image](https://user-images.githubusercontent.com/104229/94062516-13d1a300-fde7-11ea-984a-e23af64bb605.png)

The green triangle is generated using the old code, while the black dots and blue lines are added using my angle calculation. This is for a point that the old code considers "inside" the tolerance, and the new code outside. So what's going on? I added the red lines, calculated using

    a = np.radians(DV.tolerance)
    ax.plot([0, dists[x] * np.cos(a)],
           [0, dists[x] * np.sin(a)], color="red")
    ax.plot([0, dists[x] * np.cos(-a)],
           [0, dists[x] * np.sin(-a)], color="red")

(I first tried half the tolerance, and that gives an even narrower angle, obviously). Why don't they agree with the green ones?

So I looked at the _compass function, more specifically this line:
https://github.com/mmaelicke/scikit-gstat/blob/master/skgstat/DirectionalVariogram.py#L760

This can't be correct, can it? It should read `r = xmax / cos(a)`, not `r = xmax / a`, right?

redhog:
Update: Changing this to 

    r = xmax / np.cos(a)

makes the new and old methods agree 100% about DV._direction_mask()

mmaelicke:
You are absolutely right: 

https://github.com/mmaelicke/scikit-gstat/blob/e09a65047bbd5430e55d26c967be4ddcfe41849e/skgstat/DirectionalVariogram.py#L738

In the docstring, I am stating that *r = xmax / cos(a)* with *a = alpha / 2*, but in the implementation I omitted the cos. That means, the unittests test against a wrong result. Sorry, I was really in a hurry at that time and didn't double check the unittest results before implementing them.

redhog:
Awesome to have that figured out. I'll make a PR with a fix to that and the unittests, and then we can check the direction PR using those fixed unittests...

PR #52, opened 2020-09-24
Angle math bugfix

redhog:
Fixed two issues

* For compass search area, the r is now correctly calculated as `r = xmax / np.cos(a)`, see https://github.com/mmaelicke/scikit-gstat/issues/51
* For triangle search area, the shape is now correctly calculated when h > xmax (by generating the same shape compass would)

### DIFF
--- a/skgstat/DirectionalVariogram.py
+++ b/skgstat/DirectionalVariogram.py
@@ -641,6 +641,9 @@ class DirectionalVariogram(Variogram):
         # get the maximum x coordinate in the current representation
         xmax = np.max(local_ref[:, 0])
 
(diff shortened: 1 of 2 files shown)


mmaelicke:
I think things are working fine. We can merge and pull it into #50, right?

Maybe we need another unittest, that calculates the angles for only 5 points or something, to verify it by pen and paper. But that would go into another issue, just thinking out loud...

redhog:
I think a mini-dataset checked item-by-item on paper would be a great idea, but yes, I think that should be a different ticket.

I already merged this into the direction branch / pr, just not into master.

PR #54, opened 2020-10-02
Crosskriging

redhog:
* Do not merge into the direction branch. Rather, merge https://github.com/mmaelicke/scikit-gstat/pull/50 first, then change the base of this PR to master and merge.

This adds two parameters to OrdinaryKriging: coords and values, allowing the user to interpolate another dataset than the one used to generate the variogram.


redhog:
Come to think of it, maybe these should be parameters to transform? Not sure what's more "scikity"...

PR #57, opened 2020-11-24
Distance calc speedup

redhog:
Don't loop in python, let pdist do the job


mmaelicke:
Oh wow, feels almost embarrassing to still have code like this online...

Thanks a lot!

redhog:
NP :) Oh, and there is a similar thing going on with the kriging, see https://github.com/mmaelicke/scikit-gstat/issues/58

Issue #58, opened 2020-11-25
Optimize distance function in kriging


redhog:
When krigin, a distance function is called for each point to calculate its distance to every point in the variogrammed dataset. This might be necessary if the use has supplied their own distance metric, but if it's one of the built in scipy, we can use scipy.spatial.distance.cdist on all points instead, and if it's euclidean, we can even use scipy.spatial.cKDTree, see ttps://iliauk.wordpress.com/2016/02/16/millions-of-distances-high-performance-python/ , which would automatically cull the points farther away than the range without actually doing a distance calculation on them.

- [X] Optimize [observation selection](https://github.com/emerald-geomodelling/upstream-scikit-gstat/blob/2faae465b9a42c2b00e302157c44954bc510713c/skgstat/Kriging.py#L380)
- [x] Optimize [kriging equation creation](https://github.com/emerald-geomodelling/upstream-scikit-gstat/blob/2faae465b9a42c2b00e302157c44954bc510713c/skgstat/Kriging.py#L396)

mmaelicke:
I commented in #60 ...

mmaelicke:
@redhog was this resolved by #60 ?

redhog:
Only half-way. I added the two subtasks above...

mmaelicke:
OK, great! Then I'll leave this open.

redhog:
The idea I have is to generate a pdist matrix for _all_ observation points at the top, then select the sub-matrix for the nearby points instead of calling self.dist().
This might even be worth allowing "short circuit" for between Kriging and Variogram, that is, allow you to send in a distance matrix alongside with coordinates and values.

redhog:
As far as I can see this is solved in https://github.com/mmaelicke/scikit-gstat/blob/master/skgstat/Kriging.py#L415

PR #60, opened 2020-11-25
Use scipy.spatial.distance.cdist to calculate all distances at once

redhog:
Part of https://github.com/mmaelicke/scikit-gstat/issues/58

### DIFF
--- a/skgstat/Kriging.py
+++ b/skgstat/Kriging.py
@@ -10,6 +10,7 @@ from scipy.spatial.distance import squareform
 from scipy.linalg import solve as scipy_solve
 from numpy.linalg import solve as numpy_solve, LinAlgError, inv
 from multiprocessing import Pool
+import scipy.spatial.distance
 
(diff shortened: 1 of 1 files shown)


mmaelicke:
That's the way better approach. Do you want to merge it like that, or propose an implementation of `scipy.spatial.cKDTree` here as well, as described in #58 ?
If so, are there any downsides of using cKDTree (especially on small datasets) ? I don't know that method, but seems really helpful here.

redhog:
Please merge as is, and I'll make another PR for the cKDTree.

The only downside is that cKDTree can only work in euclidean space, so there will have to be an if-case handling that vs all other distance metrics. But the speedup (and RAM saving!) for larger datasets where range << max(coords) - min(coords), is large enough that it's worth it anyway. From what I know / have seen it doesn't have any performance problem for smaller datasets.

mmaelicke:
merged

PR #61, opened 2020-11-25
Separate class to store variogram results in

redhog:
This class lets you store variogram result parameters (sill, range, nugget as well as metrics used etc) without the coordinates and values, making the pickle size much smaller. It also allows you to re-add the coordinates and values, or possibly other coordinates and values for special cases.

### DIFF
--- a/skgstat/VariogramResult.py
+++ b/skgstat/VariogramResult.py
@@ -1,0 +1,62 @@
+import numpy as np
+import pandas as pd
+from .Variogram import Variogram
+
+class LagClass(object):
+    def __init__(self, size):
+        self.size = size
+
+class VariogramResult(Variogram):
+    """This class represents a calculated variogram without the original
+    coordinates and values. This is then a much smaller object and can
+    easily be stored in a database etc.
+
+    Usage:
+
+      v = skgstat.Variogram(coordinates, values)
+      vr = VariogramResult(v) # Throw away coordinates and values (and intermediate large arrays)
+
+      ...
+
+      ok = skgstat.OrdinaryKriging(vr.with_values(coordinates, values), **kriging_args)
+    """
+    def __init__(self, variogram):
+        info = variogram.describe()
+        Variogram.__init__(self, model=info["name"], estimator=info["estimator"])
+        self.cof = variogram.cof.copy()
+        self._bins = variogram.bins.copy()
+        self.__experimental = variogram.experimental.copy()
+        self._lag_classes = [LagClass(cls.size) for cls in variogram.lag_classes()]
+    @classmethod
+    def new_from_params(cls, cof, bins=None, experimental=None, lag_classes=None, **kw):
+        self = object.__new__(cls)
+        Variogram.__init__(self, **kw)
+        self.cof = np.array(cof)
+        if bins is not None:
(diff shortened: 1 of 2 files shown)


mmaelicke:
Maybe it's just the class name proposed for this new class, but I'm not 100% happy with this enhancement. 
I would hesitate to offer a possibility to store or even use a variogram model without the context of its data. For me, a model is only useful, when I know where the parameters come from and how they were estimated. Moreover, if we make it possible to pass new data (coordinates, values) to that instance, but keep the old parameters, that can easily lead to wrong assumptions about the model, doesn't it?

If it's just about storing parameters somewhere, there  are 
https://github.com/mmaelicke/scikit-gstat/blob/bab68cf32551595bf66430e0354dbafba44d3573/skgstat/Variogram.py#L1409

and 
https://github.com/mmaelicke/scikit-gstat/blob/bab68cf32551595bf66430e0354dbafba44d3573/skgstat/Variogram.py#L1453

that can already be used to return results. We can of course extend the output of these functions to bins edges etc., if needed.
So maybe @redhog you could give me some more insight, why it would be important to have a variogram without data and intermediate results as an object of its own? Then I maybe understand better what the purpose of this PR is.

If it's just about decreasing the in-memory size of the instance passed to `OrdinaryKriging`, another approach would be to make the Kriging class accept coordinates, values, and variogram parameters directly...
Or do I miss something here?

redhog:
We have large datasets with multiple measurements stored in a PostGIS database. In our workflow, we generate variograms as a separate task where the user loads some dataset of measurements, selects some parameters and then reviews the plots of a set of variograms, one for each measurements, and when happy saves the variograms to the database (with a name). Multiple variograms (using e.g. different distance metrics or estimator) can be stored for the same measurement and dataset.

In later tasks, the user loads one of the variograms for each of their currently loaded measurements, selects a polygon and raster resolution, or set of positions, and interpolates the measurements to those positions.

We store the variograms by pickling them and storing the pickled bytes to a row in the database. We later load a dataset from the database, as well as associated variogram pickles that we unpickle and insert the values and coordinates into to do kriging. Storing the data twice - both as individual database rows, and inside a pickle of the variogram in one row, is actually impossible - the database driver crashed when I tried. It would also mean that we'd have to store it multiple times if the user generated multiple variograms with different metrics etc for the same measurement.

redhog:
The idea of this class is that it encapsulates all the data that OrdinaryKriging needs, except for coordinates and values.

mmaelicke:
OK I see. That makes the case clearer for me. 
I personally would still like to go for the solution to enhance the describe() output. If I got you right, that will serve your needs and make me happy as well.
Would that be fine for you?

redhog:
If we can modify OrdinaryKrigin as per above, then that's actually better than my solution, as the describe() output can be stored as json.

However, ideally, all parameters to Variogram.__init__() should be output too, so that it is clear how the variogram got created (this would typically be shown in a info-dialog somewhere when you select a stored variogram).

mmaelicke:
100% agree. Do you want to create an issue on this, or just commit the proposed changes here?

redhog:
Hm, let's create an issue

PR #62, opened 2020-11-25
cKDTree optimization

redhog:
Closes https://github.com/mmaelicke/scikit-gstat/issues/58


skgstat/Kriging.py
```diff
@@ -389,16 +395,23 @@ def _krige(self, idx):
         dists = self.transform_dists[idx,:]
         
         # find all points within the search distance
-        idx = np.where(dists <= self.range)[0]
+        if isinstance(dists, scipy.sparse.spmatrix):
+            idx = np.array([k[1] for k in dists.keys()])
```
redhog, line 399: For non-sparse datasets this might actually be a performance bottleneck, and we should use toarray() or somesuch solution to speed it up.

redhog:
Heyas @mmaelicke ! Would you have time to review this?

Note: There is a new flag "sparse" added. If set to false, all distances are calculated using pdist, which is the fastest way, at least for semi-small datasets, but quickly eats a lot of ram (N*M, where N is number of points kriged to and M is number of points kriged from). If set to true, distances are calculated using ckDTree only for points withing range, and stored in a sparse matrix. This takes considerably less storage, but it slightly slower for lookup, so for smaller datasets this is a disadvantage. For larger datasets this saves your machine from swapping, which would quickly lower your performance.

mmaelicke:
@redhog , awesome, thank a lot! 
Yeah, I will review it. I'll do my very best to do it this evening.

redhog:
Maybe the default for sparse needs to be False... Hm...

skgstat/Kriging.py
```diff
@@ -30,6 +30,11 @@ class IllMatrixError(RuntimeWarning):
 def inv_solve(a, b):
     return inv(a).dot(b)
 
+def sparse_dok_get(m, fill_value=np.NaN):
```
mmaelicke, line 33: Can we rename it to `_sparse_dok_get`, as this is not intended to be exported from the module, right?
And maybe a one-liner docstring might be helpful

skgstat/Kriging.py
```diff
@@ -41,7 +46,8 @@ def __init__(
             precision=100,
             solver='inv',
             n_jobs=1,
-            perf=False
+            perf=False,
+            sparse=True
```
mmaelicke, line 50: I would plead for a `False` as the default value here. First, a `True` would change the current behavior. Second, I feel like the use cases in which `sparse=False` makes more sense occur more often. But I might be wrong in that feeling ;) 
We could also keep it `False`for now and turn it to `True` with the next minor version shift, along with a warning that will be raised until then.

skgstat/Kriging.py
```diff
@@ -386,24 +405,38 @@ def _krige(self, idx):
...
 
         if idx.size > self._maxp:
-            sorted_idx = np.argsort(dists)
-            idx = sorted_idx[np.isin(sorted_idx, idx)][:self._maxp]
-
+            if isinstance(dists, scipy.sparse.spmatrix):
+                selected_dists = dists[0, idx].toarray()[0,:]
+            else:
+                selected_dists = dists[idx]
+            sorted_idx = np.argsort(selected_dists, kind="stable")
```
mmaelicke, line 428: just out of curiosity: why stable sort here?

mmaelicke (changes requested):
Nice work, thanks a lot!
I think, we need to sort the default value for `sparse`out, then I'm completely fine with it.

skgstat/Kriging.py
redhog: Because without that I couldn't do a regression test that gave the same result with sparse/non-sparse (when point coords coincided exactly for some pair of points, which some did in my test dataset)..

skgstat/Kriging.py
redhog: Hm, that's probably a good idea. Will push a change.

skgstat/Kriging.py
redhog: Sure. It's just a utility function.

redhog:
This is superseded by https://github.com/mmaelicke/scikit-gstat/pull/68 which implements the same feature, but in a separate class that can also be used by Variogram.

mmaelicke:
> This is superseded by #68 which implements the same feature, but in a separate class that can also be used by Variogram.

So, you want to only merge #68 instead of this PR, or does #68 just replace the utility function after this PR was merged?

redhog:
Only merge the other one. Sorry for coding faster than I talked to you...

redhog:
I had the other one in parallel, but didn't want to push it due to a bug that I have now squished...

mmaelicke:
NP. I will just wait until you want me to review or merge something. Or both. :)

PR #65, opened 2020-12-01
Multiple plotting backends

mmaelicke:
With this PR, it has become possible to use [plotly](https://plotly.com/python/) instead of matplotlib for plotting. These plot objects integrate much better in web-driven applications using `scikit-gstat`.

The new sub-module `skgstat.plotting` has a function to switch backends:

```python
from skgstat import plotting

plotting.backend('plotly')
[...]

Vargriogram.plot()
```

plotly is **not** added as a dependency, so it has to be installed manually. The new code didn't change any existing behavior and  `skgstat` can be used just like before.

At the current state, `skgstat` does not make use of the most powerful features of plotly, as I am just porting the existing plots to plotly. This will be a focus of future developments, to make more of plotly's (and also matplotlib's) features and customizations available.

@redhog , I will add a demo notebook as quick as possible and then I'm curious to hear your thoughts on this.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 [run]
 omit = 
     skgstat/tests/*
+    skgstat/plotting/*
     docs/*
     setup.py
 
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,6 @@ docs/savefig
 notebooks
 .vscode
 container
+
+# project specific
+Playground.ipynb
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 0.3.1
+=============
+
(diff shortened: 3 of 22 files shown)


redhog:
Cool.
I might be able to provide some bokeh based plotting routines if you're interested... But the code would need some serious cleanup first :P

mmaelicke:
> Cool.
> I might be able to provide some bokeh based plotting routines if you're interested... But the code would need some serious cleanup first :P

Yeah, always!
In fact I rolled a dice whether to implement bokeh or plotly first :). Personally, I use plotly more often, because I use dash on a very frequent basis. 
But the new code is designed with more that just matplotlib and plotly in mind. Thus it should be quite easy to implement bokeh once the plotly stuff is finished.

Issue #67, opened 2020-12-02
Restructuring for optimization


redhog:
I've been thinking a bit about the class structure and how things could/should be structured to make more optimizations possible, esp in the case where you have multiple variograms for different measurements at the same positions:

This is not intended as a "let's hack this now", but as a discussion starter to see what you think, and maybe there's a totally different better way to structure the API.

We could have a class that models a set of positions together with their distance metric:

```
pos = MetricSpace(coordinates, metric="euclidean")
new_pos = MetricSpacePair(pos, new_coordinates)

v1 = Variogram(pos, values1, ...)
ok1 = OrdinaryKriging(pos, values, v1.describe())
v2 = Variogram(pos, values2, ...)
ok2 = OrdinaryKriging(pos, values, v2.describe())
...

interp_values1 = ok1.transform(new_pos)
interp_values1 = ok2.transform(new_pos)
...
```

The idea here is that MetricSpace calculates the (possibly sparse) distance matrix once of all points in coordinates to all other points in coordinates, regardless of how many variograms or krigings you use it for, and MetricSpacePair does the same for the distance matrix of all points in coordinates and all points in new_coordinates.

mmaelicke:
Thanks for the suggestions. 

So far I didn't put too much effort and thinking into the kriging class, so improvements are welcome. I think if we design the `MetricSpacePair` class wisely, the current code is not too far away from integrating the behavior you describe above. Given that the `OrdinaryKriging` class is refactored on input attributes as described in #63 , it's only the `OrdinaryKriging.transform` method that needs some rework. Right now, it needs N 1D-arrays, one for each coordinate dimension. 
So either the transform method signature is changed to also accept a `MetricSpacePair`, or the `MetricSpacePair` needs a function to export the new positions in the correct format. 
The advantage of accepting a `MetricSpacePair` instance would be, that `OrdinaryKriging` could pass around an internal flag that it received a `MetricSpacePair` instance and prevents itself from re-calculating distances, as `MetricSpacePair` would hold a distance matrix for the new positions, right?

That would be cool, as it is a real improvement, but `OrdinaryKriging` would still accept a single variogram and calculate the distance matrix itself, if no `MetricSpacePair` is given.

---

A few more thoughts that might be interesting here:

1. In the past, there was always the idea to write a proper interface between `skgstat.Variogram` and [pykrige](https://pypi.org/project/PyKrige/). 

an experimental interface function is here:
[...]

PR #68, opened 2020-12-17
Metric space

redhog:
First steps towards extracting all distance metrics representation into a separate object as outlined in https://github.com/mmaelicke/scikit-gstat/issues/67


redhog:
Hi @mmaelicke !

Please take a look. There is one issue with this code: I'm not sure how to serialize the parameters of a harmonized model to the describe() output in such a way that the model function could be recreated in Kriging. I haven't worked with harmonized models, so I'm not sure how it's supposed to work. Any thoughts?

skgstat/Variogram.py
```diff
@@ -996,22 +986,41 @@ def fitted_model(self):
...
+                cof.append(kw["smoothness"])
+            if "shape" in kw:
+                cof.append(kw["shape"])
+            if kw["nugget"] != 0:
+                cof.append(kw["nugget"])
+
+        harmonize = False
+        if not callable(model):
+            if model == "harmonize":
+                # FIXME: How do we serialize a harmonized model in describe()?
```
redhog, line 1007: This is where I don't really know what to do...

mmaelicke: I will have a look at it as soon as possible.

mmaelicke: Hey @redhog, I don't find the time right now to solve this (lockdown here, kids at home etc.)...
Would it be a possible pathway for you to **not** include the parameters of harmonized models for now? Then, the Kriging class would raise a `NotImplementedError` if model params are missing. Then, we could merge this PR and open an issue concerning the harmonized model, to keep it on the agenda.
I just don't want to hold you up any longer, sry about that.

redhog: I don't need harmonized models myself, so for me that'd work. However, it would break the workflow for anyone who do. If you think that's not too big a deal, we could merge anyway. Alternatively I guess I could add a hack workaround: Copy the model function (not just name) here: https://github.com/mmaelicke/scikit-gstat/blob/cf0751b5c45b9acaacb7cb2086c42fa881d440a7/skgstat/Kriging.py#L105

mmaelicke: Then I would go for hacky workaround along with an issue opened to fix it at any later point in time.

redhog:
Heyas @mmaelicke!
Just wanted to let you know I'm working on this again, and have fixed the harmonized model problem.
I also updated the `Variogram` model to use `MetricSpace`, so that when you set `maxlag` it can take advantage of this to not consider all pairs.
I do still have some faulty unittests, mostly because they assume `coords` is a `ndarray`, not a `MetricSpace`.
Btw, this is the shape of my current test dataset:

```
                   X             Y
count   16887.000000  1.688700e+04
mean   105876.886644  1.256407e+06
std       162.765830  2.139842e+02
min    105374.220678  1.256050e+06
25%    105757.755000  1.256232e+06
50%    105904.848719  1.256376e+06
75%    106002.500000  1.256596e+06
max    106212.660078  1.256800e+06

```

And I use a maxlag of 50...

mmaelicke:
Hey, thank man! Looks nice! 
There is a lot of code outside of skgstat that actually assumes `coordinates` to be a ndarray. Maybe it's a good idea to store the MetricSpace as a new attribute and let `Variogram.coordinates` return the coords just like before, to not break existing code. 
It should be a property function anyway, so implementing it like this should be straightforward. 
I can have a more detailed look on the code, if you want me to. I just quickly scanned the changed files, to be honest.

redhog:
Sounds like a good plan, and probably fixes some of the tests. I'll do that.

redhog:
`Variogram.coordinates` is actually already a ndarray (it's a property that points to `self._X.coords`)

redhog:
@mmaelicke

Heyas! I think this is ready for merging now. Added some unit tests.
In general for testing Variogram(), it's good to have test data that is actually spatially correlated. I hacked that together for testing maxlag, but it's a bit of a hack...

mmaelicke:
Nice, I'll have a look at it somewhere this week. 
We're still losing some test coverage. But yeah I agree, there should be some tests that use data that is actually spatial correlated. 
BTW., I merged a new property today, that returns the spatial dimensionality, which is right now relying on `_X`. That will need fixing after this merge, as well. 
@redhog, did you check if the examples in the docs are still compiling? There is a lot of quite old code in there, possibly some of the examples are using `_X` directly and need to call `Variogram.coordinates` to get the ndarray of coordinates. That would be a useful update anyway. This is something that I can do myself..

Issue #78, opened 2021-03-22
GStools interface bug


MuellerSeb:
https://github.com/mmaelicke/scikit-gstat/blob/664eb89a8d578ff27c7e820e2edbd34ed3ee3ac1/skgstat/interfaces/gstools.py#L44

The referenced line is most probably not doing anything useful, since the overridden `variogram` method doesn't make use of the internal attributes `var`, `nugget`, `len_scale` or any aditional hyperparameter.

This should be fixed by a better conversion function in the future.

mmaelicke:
I am not sure if the interface is doing something useful at all... :)
I think we can completely replace this with a new interface. I give you a call in the next couple of days. 
On mid-term, I want to replace the whole `interface` submodule anyway. 

I'll keep this open as a reminder and to build up some pressure on myself.

MuellerSeb:
Cross-referencing this issue here: https://github.com/GeoStat-Framework/GSTools/issues/92

PR #79, opened 2021-04-07
GSTools: add interface to map skgstat models onto GSTools CovModels

MuellerSeb:
This is a first Draft.

One can now do the following:
```python
import numpy as np
import matplotlib.pyplot as plt
from skgstat import Variogram

plt.style.use('ggplot')

xx, yy = np.mgrid[0:0.5 * np.pi:500j, 0:0.8 * np.pi:500j]
np.random.seed(42)
_field = np.sin(xx)**2 + np.cos(yy)**2 + 10
np.random.seed(42)
z = _field + np.random.normal(0, 0.15, (500, 500))
plt.imshow(z, cmap='RdYlBu_r')

np.random.seed(42)
coords = np.random.randint(0, 500, (300, 2))
values = np.fromiter((z[c[0], c[1]] for c in coords), dtype=float)
V = Variogram(coords, values, model="stable")
fig = V.plot()

gs_model = V.to_gstools()
gs_model.plot(ax=fig.axes[0], x_max=gs_model.len_scale)

print(V)
print(gs_model)
```
And gets:
```bash
stable Variogram
----------------
Estimator:         matheron
Effective Range:   628.27
Sill:              0.57
Nugget:            0.00

Stable(dim=2, var=0.57, len_scale=6.28e+02, nugget=0.0, anis=[1.0], angles=[0.0], alpha=2.0)
```
skgstat_to_gstools

As you can see, the plotted variograms perfectly match.

See: 
- https://github.com/mmaelicke/scikit-gstat/issues/78
- https://github.com/GeoStat-Framework/GSTools/issues/92


mmaelicke:
Thanks Sebastian! 
Great, I'll have a look as soon as possible. Looks really cool! Thank you so much.

skgstat/interfaces/gstools.py
```diff
@@ -1,46 +1,56 @@
...
+        import gstools as gs
+    except ImportError:
+        raise ImportError("skgstat_to_gstools: GSTools needs to be installed.")
+    kwargs.setdefault("dim", variogram.coordinates.ndim)
+    describe = variogram.describe()
+    name = describe["name"]
+    if name not in MODEL_MAP:
+        raise ValueError("skgstat_to_gstools: model not supported: " + name)
+    gs_describe = MODEL_MAP[name]
+    gs_describe.setdefault("rescale", 1.0)
```
MuellerSeb, line 43: This sould be `"rescale", None`

skgstat/interfaces/gstools.py
```diff
@@ -1,46 +1,56 @@
...
+)
+
+
+def skgstat_to_gstools(variogram, **kwargs):
+    """Instantiate a corresponding GSTools CovModel."""
+    try:
+        import gstools as gs
+    except ImportError:
+        raise ImportError("skgstat_to_gstools: GSTools needs to be installed.")
+    kwargs.setdefault("dim", variogram.coordinates.ndim)
```
MuellerSeb, line 37: Is that really the spatial dimension? Just copied this line from the old interface.

mmaelicke: Probably not. `Variogram.coordinates` points to a numpy.ndarray of shape `(n_obs, n_dims)`

I.e. for 100 3D coords the shape will be `(300, 3)`. Thus the `ndim` would be `2` (I think) and `variogram.coordinates.shape[1]` is the actual spatial dimension, right?

Thanks for pointing out.

MuellerSeb: You mean `(100, 3)` for 100 3D values, right? OK, for 1D I need to assure that there is an array with 2 axis.
```python
np.array(variogram.coordinates, ndmin=2).shape[1]
```
Or you could already convert this on creation for the Variogram class. Maybe you could also create an attribute holding the dimension doing this internally, so we just use `variogram.dim` here.

mmaelicke: Yeah `(100,3)`.
I'll go for the `dim`. If I am not mistaken, 1D arrays are transformed to (x, 2) anyway, but `Variogram` takes the property approach everywhere anyway, so having a `Variogram.dim` makes much sense and looks quite clean.

MuellerSeb: AFAICT, it is not transformed inplace, so `variogram._X` will be a 1D array if given as 1D array. So you could go with:
```python
@property
def dim(self):
    return np.array(self._X, ndmin=2).shape[1]
```
I don't know what happens for `coordinates=None`.

mmaelicke: Just realized the same thing. Right now, @redhog is writing a class to wraps input coordinates internally, that handles sampling for a very large input dataset.
Thus, the current implementation of `_X` will change soon. Therefore, using properties like `coordinates` and `dim` is the best way to produce the same output when the implementation under the hood changes.

For `Variogram` passing coordinates is mandatory. Thus, `coordinates=None` is kind of wrong usage, but I think I will add an `AttributeError` if coordinates is none.

mmaelicke: Found it: 

https://github.com/mmaelicke/scikit-gstat/blob/a2bf354a4b3243453c11a6b3154709390858f89c/skgstat/Variogram.py#L1395

Oh boy, not a super good style to hide that on line 1395....
However, I'll go for a `dim` property that handles the dimensionality correct and make sure that it output doesn't change no matter what we implement under the hood...

mmaelicke: `Variogram.dim` is now implemented with version `0.4.3`.

MuellerSeb: > Found it:
> 
> https://github.com/mmaelicke/scikit-gstat/blob/a2bf354a4b3243453c11a6b3154709390858f89c/skgstat/Variogram.py#L1395
> 
> Oh boy, not a super good style to hide that on line 1395....
> However, I'll go for a `dim` property that handles the dimensionality correct and make sure that it's output doesn't change no matter what we implement under the hood...

Saw that as well. But `self._X` is not overwritten, so this shouldn't harm the proposed `dim` attribute.

skgstat/interfaces/gstools.py
MuellerSeb: Actually it is fine.

mmaelicke:
@MuellerSeb I'll come to this PR soon, I am currently working on finally merging #84 after it's predecessor is open since December...
So impolite, but I have not forgotten about this...

skgstat/interfaces/gstools.py
```diff
@@ -1,46 +1,87 @@
...
+
+    Parameters
+    ----------
+    variogram : :any:`Variogram`
+        Scikit-Gstat Variogram instance.
+    **kwargs
+        Keyword arguments forwarded to the instantiated GSTools CovModel.
+        The default parameters 'dim', 'var', 'len_scale', 'nugget',
+        'rescale' and optional shape parameters will be extracted
+        from the given Variogram but they can be overwritten here.
```
mmaelicke, line 45: I am just wondering if this is actually helpful. On the one hand I think it could be counter-intuitive to fit a skg.Variogram that shows i.e. a specific sill on the plots, but then overwrite the 'var' argument, which effectively makes it another Variogram. On the other hand, it will be the user who is passing in the arguments explicitly, so I guess he has good reasons to do so.
@MuellerSeb, what do you think about that?

skgstat/interfaces/gstools.py
```diff
@@ -1,46 +1,87 @@
...
+    return np.power(3, 1 / describe["shape"])
 
 
-def gstools_cov_model(variogram, **kwargs):
-    """GSTools Interface
+def matern_rescale(describe):
+    """Get GSTools rescale parameter from sk-gstat matern model description."""
+    if 0.5 < describe["smoothness"] < 10.0:
+        return 4.0
+    return 6.0
```
mmaelicke, line 14: This is a reminder to myself: I need to change the Matern model and always use 6.0 here.

skgstat/interfaces/gstools.py
```diff
@@ -1,46 +1,87 @@
...
-        kwargs['dim'] = variogram.coordinates.ndim
-
-    # Create the instance
-    model = VariogramModel(**kwargs)
-    model.fit_variogram(variogram.bins, variogram.experimental)
-
-    return model
+    try:
+        import gstools as gs
+    except ImportError as exc:
```
mmaelicke, line 63: I personally always use just `e` as the exception object here. So I will rename `exc` here.

mmaelicke (approved):
@MuellerSeb thank you so much. This is a great PR and really, really helpful.
There is only one comment, that you are welcome to reply to. I will merge this into a `gstools_interface` branch on my scikit-gstat and add some comments to the code, so that I will understand it, as well.
Before I finally merge into main, I want to fix my Matern model again, but I take care of fixing the rescale parameter then. 
Thanks again for your contribution!

PR #82, opened 2021-04-09
Adding `get_empirical`function

### LINKED ISSUE #80: Output function for experimental Variogram
Implement a new return function for `Variogram`, that returns the bin edges (or centers) along with the current experimental data. This is helpful to align skgstat with the `gstools.vario_etimate` function.

Could look like:

```Python
Variogram:
    def get_empirical(self, bin_center=False) -> Tuple[np.ndarry, np.ndarray]:
        if bin_center:
            bins = ...
        else:
            bins = self.bins
        
        return (bins, self.experimental)
```

mmaelicke:
This closes #80 but is still a draft. Nothing implemented yet.

- [x] add function
- [x] add test

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,9 +2,23 @@
 Changelog
 =========
 
+Version 0.4.2 [WIP]
+===================
+- [Variogram] :func:`bins <skgstat.Variogram.bins>` now cases manual setted bin edges automatically
+  to a :func:`numpy.array`.
+- [Variogram] :func:`get_empirical <skgstat.Variogram.get_empirical>` returns the empirical variogram.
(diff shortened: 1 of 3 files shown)


mmaelicke:
@MuellerSeb, one thing that came to my mind is, that there are binning functions in skgstat that yield uneven bin edges/centers. Something like `[2, 4, 18, 40]` as an extreme example would be possible. Is that a problem if used in `gstools`?

The `Variogram` is aware of the kind of procedure and could raise a warning or print a hint if necessary. 
In case you are fine with the function, I can merge and release `0.4.2`

MuellerSeb:
GSTools can handle uneven bins, so this is fine.

MuellerSeb:
For reference: https://github.com/GeoStat-Framework/GSTools/issues/92

PR #84, opened 2021-04-12
Metric space

redhog:
Continuation of https://github.com/mmaelicke/scikit-gstat/pull/68 but within this repo


redhog:
@mmaelicke Do you think we can get this merged before the weekend?

mmaelicke:
> @mmaelicke Do you think we can get this merged before the weekend?

Yes. Definitely. I'll try to add some tests today and tomorrow. If I don't manage to do so, we can merge tomorrow anyway...

redhog:
Awesome!

skgstat/Kriging.py
```diff
@@ -88,14 +91,25 @@ def __init__(
             If True, the different parts of the algorithm will record their
             processing time. This is meant to be used for optimization and
             will be removed in a future version. Do not rely on this argument.
+        sparse : bool
+
```
mmaelicke, line 95: Can't we get this information from the Variogram? The internal metric space should know, right?
Or is this referring to something else that is spase? 
Maybe one sentence in the docstring would be helpful

skgstat/MetricSpace.py
```diff
@@ -0,0 +1,120 @@
+import scipy.spatial
+import scipy.spatial.distance
+import scipy.sparse
+import numpy as np
+
+def sparse_dok_get(m, fill_value=np.NaN):
+    mm = np.full(m.shape, fill_value)
+    for (x, y), value in m.items():
+        mm[x,y] = value
+    return mm
```
mmaelicke, line 10: This function should be documented and tested. Especially if it is public. If it's not, we could add a underscore-prefix.
I, myself, don't get what it does, so a docstring would be nice to have anyway

skgstat/MetricSpace.py
```diff
@@ -0,0 +1,120 @@
...
+
+class DistanceMethods(object):
+    def find_closest(self, idx, max_dist = None, N = None):
+        """Find the (N) closest points (in the right set) to the point with
+        index idx (in the left set).
```
mmaelicke: Is max_dist the same as `maxlag`? 
What is the left and right set here referring to?

skgstat/MetricSpace.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        assert self.max_dist is None or max_dist is None or max_dist == self.max_dist, "max_dist specified and max_dist != self.max_dist"
```
mmaelicke, line 34: Can we break this down into an if?

skgstat/VariogramResult.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+import pandas as pd
+from .Variogram import Variogram
+
+class LagClass(object):
+    def __init__(self, size):
+        self.size = size
+
+class VariogramResult(Variogram):
```
mmaelicke, line 9: Is it really a good idea to inherit from Variogram here? 
I think I don't really get why this is necessary.  I imagine, you can get somehow conused if a class has methods like fit and so on, that don't actually do anything. What do you think?

skgstat/VariogramResult.py
```diff
@@ -0,0 +1,62 @@
...
+        self.cof = np.array(cof)
+        if bins is not None:
+            self._bins = bins
+        if experimental is not None:
+            self.__experimental = experimental
+        if lag_classes is not None:
+            self._lag_classes = lag_classes
+        return self
+    def fit(self, *arg, **kw):
+        pass
```
mmaelicke, line 43: I would rather raise an exception that a VariogramResult can not be fitted.

skgstat/tests/test_variogram.py
```diff
@@ -15,10 +15,51 @@
     PLOTLY_FOUND = False
 
 from skgstat import Variogram
+from skgstat import OrdinaryKriging
 from skgstat import estimators
+from skgstat import models
 from skgstat import plotting
 
 
+class TestSpatiallyCorrelatedData(unittest.TestCase):
```
mmaelicke, line 24: NIce. Thanks!
Is your idea to rather replace all the usages of not-correlated samples in the existing unittests with this class or to add additional new tests to the current test suite?
Then we could put the spatially correlated into a test file of its own, as this one is already quite big and confusing.

mmaelicke (changes requested):
@redhog thank you for all this great work and improvements. Thank you also very much for all the typo and minor fixed included. 
However, I have one issue with the VariogramResult class. Is it already finished? then it could need some documentation and test coverage. Otherwise that might go into a PR of its own.
Secondly there is one thing about coordinates and values passed to Kriging, that we should resolve, before merging.
Thank you again

skgstat/tests/test_variogram.py
redhog: I think given that I use a Variogram & Kriging to generate the data, it's not good to have it in the main class - if _anything_ is broken, every single test might fail in the setUp() function.

But we could of course generate the data some other way, or pre-generate it and store it in a csv file or something.

redhog: But for now, I just thought sticking new tests in this additional class was the simplest/fastest path to being able to test stuff using correlated data....

skgstat/Variogram.py
```diff
@@ -840,13 +847,14 @@ def use_nugget(self, nugget):
...
-        return self._dist_func_name
+        return self._X.dist_metric
 
-    def _dist_func_wrapper(self, x):
-        if callable(self._dist_func_name):
-            return self._dist_func_name(x)
+    @classmethod
+    def wrapped_distance_function(cls, dist_func, x):
+        if callable(dist_func):
+            return dist_func(x)
```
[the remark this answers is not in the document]
redhog: So dist_func can either be one of the values that scipy.spatial.distance.pdist accepts for its metric parameter, or it's a callable function that behaves like scipy.spatial.distance.pdist(X), returning a diagonal distance matrix, but with distances calculated using some custom metric.

skgstat/Variogram.py
```diff
@@ -42,18 +43,19 @@ def __init__(self,
...
 
-        Note: The directional variogram estimation is not re-implemented yet.
-        Therefore the parameters is-directional, azimuth and tolerance will
-        be ignored at the moment and can be subject to changes.
-
         Parameters
         ----------
-        coordinates : numpy.ndarray
+        coordinates : numpy.ndarray, MetricSpace
+            .. versionchanged:: 0.5.0
```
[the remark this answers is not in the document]
redhog: Sounds good to me

skgstat/Variogram.py
```diff
@@ -42,18 +43,19 @@ def __init__(self,
...
+                now accepts MetricSpace
             Array of shape (m, n). Will be used as m observation points of
             n-dimensions. This variogram can be calculated on 1 - n
             dimensional coordinates. In case a 1-dimensional array is passed,
             a second array of same length containing only zeros will be
             stacked to the passed one.
+            For very large datasets, you can use the MetricSpace class, that
+            will only calculate distances within the maximum lag in a 
+            sparse matrix.
```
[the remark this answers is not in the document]
redhog: It is semi-correct.

I'd say something like:


For very large datasets, you can set maxlag to only calculate distances within the maximum lag in a sparse matrix. Alternatively you can supply a MetricSpace (optionally with a max_dist set for the same effect). This is useful if you're creating many different variograms for different measured parameters that are all measured at the same set of coordinates, as distances will only be calculated once, instead of once per variogram.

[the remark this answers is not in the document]
redhog: A similar comment might be useful in the parameter doc for `Kriging`.

skgstat/Variogram.py
```diff
@@ -211,8 +213,27 @@ def __init__(self,
...
+                    coordinates,
+                    np.zeros(len(coordinates))
+                ))
+                self._1d = True
+
+            # handle maxlag for MetricSpace
+            _maxlag = maxlag if maxlag and not isinstance(maxlag, str) and maxlag >= 1 else None
+            coordinates = MetricSpace(coordinates.copy(), dist_func, _maxlag)
+        elif self.dist_func != coordinates.dist_metric:
+            raise AttributeError("Distance metric of variogram differs from distance metric of coordinates")
```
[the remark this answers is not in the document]
redhog: Nah, I don't have a strong opinion on this :)

skgstat/Variogram.py
```diff
@@ -1383,33 +1392,39 @@ def fitted_model(self):
...
 
-        # get the function
-        func = self._model
-
-        if self._harmonize:
-            code = """model = lambda x: func(x)"""
+        return self.fitted_model_function(self._model, self.cof)
+
+    @classmethod
+    def fitted_model_function(cls, model, cof=None, **kw):
```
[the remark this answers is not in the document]
redhog: I it's called from `Kriging`, and is supposed to accept all arguments that `describe()` outputs, as in 

model = Variogram.fitted_model_function(**.v.describe())

skgstat/tests/test_variogram.py
mmaelicke: I also tĺike the idea of rather adding stuff more than replacing the existing. For me, it's also the CSV the way to go, but that is nothing that has to be solved in this PR and is on my agenda anyway. Therefore I'll would suggest of keeping the stuff just like it is and mark this question as resolved. Agree?

skgstat/MetricSpace.py
redhog: Yes it is, but for the limited usage of distance calculations, i.e. the usage within this class.
Left & right are applicable to MetricSpacePair(), where you have two sets of points, left being transform points and right being observations: https://github.com/mmaelicke/scikit-gstat/blob/metric-space/skgstat/Kriging.py#L295 : `MetricSpacePair(self.transform_coords, self.coords)`

skgstat/MetricSpace.py
redhog: Sure :)

skgstat/VariogramResult.py
redhog: Maybe yeah. Given the new describe(), maybe this class isn't that useful anymore even... It's not required by any of the rest of the code, and I guess I can reengineer the stuff I have that needs it...

skgstat/Variogram.py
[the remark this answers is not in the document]
mmaelicke: I see, I tend to forget the level of wrapping things, that is used in some parts in scikit-gstat...
Then I am fine with this function. Makes sense

skgstat/Variogram.py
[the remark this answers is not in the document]
mmaelicke: Ah, nice, that is actually really helpful and I was thinking about building something like this anyway...

skgstat/VariogramResult.py
mmaelicke: I would always go for dropping the class. On the other hand it does not harm the existing stuff. So afterall, it's kind of your decision, I guess

skgstat/MetricSpace.py
mmaelicke: So you would argue, that `max_dist` as an argument name makes more sense here, to distinguish it from the `maxlag` parameter?

What happens if I do something like this:

```Python
ms = MetricSpace(coords, max_dist=50)
V = Variogram(ms, values, maxlag=80)
```
Do I miss something, or does the `maxlag=80` not have any effect anymore? Sorry for missing the forest through the trees...

mmaelicke:
@redhog are you currently working on more commits?

redhog:
Nope :)

mmaelicke:
@redhog, from the conversation I get that I will turn a few `assert` statements into an `if` today. What about the `VariogramResult`?  you mentioned it could be dropped... 
Can I go forward and do that to finally merge today, or is there something that we need to take care of / discuss?

Finally, just for my understanding, a reply to https://github.com/mmaelicke/scikit-gstat/pull/84#discussion_r614018206 would be great, so that I understand your code as well :)

Anything you would like to add?

Thank you man, I am really looking forward to use these nice code extensions.

skgstat/MetricSpace.py
redhog: Yeah, but it's not super important that they're different.

There is an assert that would flag this situation. If you supply a MetricSpace to Veriogram, you can only do one of these:

* Set MetricSpace.max_dist (this will propagate to maxlag in Variogram)
* Set Variogram.maxlag (this will be honored, but won't optimize distance calculations)
* Set both to the same value

redhog:
Yes, just drop `VariogramResult`.
And no, I think any other fixes/changes I make would have to go to their own PRs - this is getting way too big...

skgstat/MetricSpace.py
mmaelicke: @redhog  what about adding a warning if maxlag and max_dist are both set, but not of same value? 
A warning will break the code, but the user is able to ignore it...

redhog: Sounds good to me.

redhog: Possibly a different warning if that's the case and maxlag > max_dist?

skgstat/VariogramResult.py
redhog: Well, drop it then :)

mmaelicke:
@redhog I am finished with my changes. But I have broken something. All 7 tests fail at the same line:

https://github.com/mmaelicke/scikit-gstat/blob/d6b05e270ec4ef9a1d4fde4242b166c9c58382c6/skgstat/MetricSpace.py#L49

I don't get it right now. Maybe I am in front of my computer for a few hours too long today.
Did I unindent something? Can you help me out?

mmaelicke:
@redhog ,
Unfortunately, I found a serious issue with this branch. I ran a notebook this morning producing some figures for my next publication. With the `metric-space` branch still active, a new variograms changed their shapes dramatically, without chaning the code. After some intensive code review, I think there is something wrong with the maxlag / max_dist settings. I tried to put an example together. 
Consider the two examples below. I am only setting a maxlag, that should not change the variogram, as I am using the largest bin as a maxlag. For the **master** branch everything is fine and the two variograms look the same:
master

For the **metric-space** branch the two variograms differ dramatically. Note: the `coordinates` is a `ndarray` of shape `(500,2)`.
metric-space
  
I am right now not sure, what is happening and you might have the better insight into the code. Maybe the grouping into lag classes is somehow not prepared for some kind of edge case. Other variograms in the very same notebook do not show this behavior.

mmaelicke:
@redhog: Edit: here is the sample dataset if you want to run / try it yourself: 
[sample.csv](https://github.com/mmaelicke/scikit-gstat/files/6334518/sample.csv)

redhog:
Thanks, was gonna ask for the sample dataset :) Will try it out and try to figure out what's boken ASAP....

mmaelicke:
@redhog  Thanks, appreciate it!
I just pushed commit https://github.com/mmaelicke/scikit-gstat/commit/0c617c5bce99ee9b87d4bc85d4d9af5cd5a6a598 to master, which includes the test data and a test that reproduces the example. You can also use that one to test a bugfix. (At this would also be another good place to put more tests that need actual spatial correlated data in the future...)

redhog:
There was some strangeness in the handling of max_dists (self.max_dist overriding max_dist to find_closest). With that fix, your new test, cherry-picked from master, passes.

mmaelicke (approved):
I'll merge and release in a minute.
Thanks @redhog, this is an awesome PR that really makes the lib better.

redhog:
:) So happy we managed to finally push it all the way!

Issue #48, opened 2020-09-22
Sampling of point pairs


redhog:
Sample the point pairs randomly to speed up handling of big datasets. Method:

* Repeat
  * Sample datapoints
  * Build & filter pairs
  * Append pairs to total list of pairs

This way, it would even be possible to include all points (but not all pairs) while keeping the complexity at O(N) rather than O(N^2).

mmaelicke:
What do you think is the number of points, at which this does really make a difference? 

The `Variogram` class is building an array, like the `pdist` function does. It's an array of `(N*2 / 2) - N` lengh for N points, like the upper triangle of a `NxN` shaped matrix containing all combinations. 
I am not so sure which parts of the class assume the array to be of that shape. If I get you right, you would randomly choose pairs over and over again, until each point is somehow incorporated, but stop before the array grows too large. I see the advantage if it is feasible to add this for large datasets **only**. For smaller ones, where the computation times are small anyway, I think it's better to be 'deterministic' with a variogram that yields the same values on every run.

redhog:
I agree about small datasets. Not sure at what size it starts to matter - possibly only really large datasets.

Right now I do a single sample before I run the directional variogram (outside of the class), but that obviously makes the results a bit unstable. I started doing this back when I was using the code using shapely - and even with the sampling my dataset took over night to compute. With the new code, it's < 30s for a 20% sample...

My idea would be to do this multiple times and average the results, "outside" of the pdist call, so that for each subsample you still build the (N*2 / 2) - N pdist-like array, but with N = sample size rather than dataset size...

mmaelicke:
Ok, that shouldn't be too complicated. If I recall my own code correct, the only function filling the array is `_calc_distances`, which is implementing the new wrapper you added.

So we could add a check here:
https://github.com/mmaelicke/scikit-gstat/blob/e09a65047bbd5430e55d26c967be4ddcfe41849e/skgstat/Variogram.py#L1015

or more precisely in front of this line:
https://github.com/mmaelicke/scikit-gstat/blob/e09a65047bbd5430e55d26c967be4ddcfe41849e/skgstat/Variogram.py#L1025

to implement something like:

```
threshold = 10000 # maybe the user can set this
if len(self._X) > threshold:
    # _x is a referebce to self._X; or _X with a stacked 0 array if _X.dim == 1
    _x = self._new_sampling_method(_x)

# now run the old line
self._dist = self._dist_func_wrapper(_x)
```

[...]

PR #87, opened 2021-04-15
Probabalistic metric space

redhog:
Point pairs can be sampled randomly. Two random subset of all points are chosen, and the distance matrix is calculated only between these two subsets. The size of each subset is set by `samples`: if < 1 it specifies a fraction of all points, if >= 1 it specifies the number of points in each subset.

Closes https://github.com/mmaelicke/scikit-gstat/issues/48

Only look at this _after_ merging https://github.com/mmaelicke/scikit-gstat/pull/84


mmaelicke:
wow it's going really fast here...
I'll try to keep up with your pace :)

redhog:
It's not at all as large a PR as it looks as it contains all of https://github.com/mmaelicke/scikit-gstat/pull/84 too, that's why that hads to be merged first. This is just a single class, ~100 lines of code max...

redhog:
But yeah, this is all stuff I'm building for a single feature in our software. For a bit of background: We have built (and are using) https://github.com/emerald-geomodelling/EmeraldTriangles to manage triangulated datasets. Among those are topography, and while that usually is delivered as a raster (geotiff) and easy to sample to the TIN (which is typically build from a grid + all measurement points), sometimes it comes as an existing TIN that we just add the new (grid + measurement) points to. In that case I wanted to use kriging on topo too, not just measurements. But for such a topo (DTM) TIN, the number of points is very large, and I ran into all the things I've been addressing in these two PRs.

redhog:
For now the scikit-gstat kriging is done outside of EmeraldTriangles, but my goal is to have a utility function in that library that uses vertice points with a non-nan value in some column as observations, kriging to all rows with nan values in the same column.

redhog:
@mmaelicke This is ready for merging now whenever you have time :)

skgstat/MetricSpace.py
```diff
@@ -243,3 +243,70 @@ def dists(self):
...
         return self._dists
+
+class ProbabalisticMetricSpace(MetricSpace):
+    """Like MetricSpace but samples the distance pairs only returning a
+       `samples` sized subset. `samples` can either be a fraction of
+       the total number of pairs (float < 1), or an integer count.
+    """
+    
+    def __init__(self, coords, dist_metric="euclidean", max_dist=None, samples=0.5):
+        self.coords = coords.copy()
```
mmaelicke, line 254: I would add a docstring here, as the user should be able to directly interact with this class, right?

skgstat/MetricSpace.py
```diff
@@ -243,3 +243,70 @@ def dists(self):
...
+    
+    @property
+    def ridx(self):
+        if self._ridx is None:
+            self._ridx = np.random.choice(len(self.coords), size=self.sample_count, replace=False)
+        return self._ridx
+    
+    @property
+    def ltree(self):
+        assert self.dist_metric == "euclidean", "A coordinate tree can only be constructed for an euclidean space"
```
mmaelicke, line 287: I will turn this into a if again ;)

skgstat/MetricSpace.py
```diff
@@ -243,3 +243,70 @@ def dists(self):
...
+        if self._ltree is None:
+            self._ltree = scipy.spatial.cKDTree(self.coords[self.lidx,:])
+        return self._ltree
+
+    @property
+    def rtree(self):
+        assert self.dist_metric == "euclidean", "A coordinate tree can only be constructed for an euclidean space"
```
mmaelicke, line 294: same here: if statement

skgstat/MetricSpace.py
```diff
@@ -243,3 +243,70 @@ def dists(self):
...
+        if self._rtree is None:
+            self._rtree = scipy.spatial.cKDTree(self.coords[self.ridx,:])
+        return self._rtree
+            
+    @property
+    def dists(self):
```
mmaelicke, line 300: This could need a docstring

skgstat/MetricSpace.py
```diff
@@ -243,3 +243,70 @@ def dists(self):
...
+        if self._dists is None:
+            max_dist = self.max_dist
+            if max_dist is None:
+                max_dist = 1 + np.sqrt(((self.coords.max(axis=0) - self.coords.min(axis=0))**2).sum())
```
mmaelicke, line 304: @redhog, is this still the intended behavior if the distance metric is not euklidean?

skgstat/Variogram.py
```diff
@@ -14,10 +14,9 @@
 from skgstat import estimators, models, binning
 from skgstat import plotting
 from skgstat.util import shannon_entropy
-from .MetricSpace import MetricSpace
+from .MetricSpace import MetricSpace, MetricSpacePair, ProbabalisticMetricSpace
```
mmaelicke, line 17: Do we actually need the import of MetricSpacePair here? I think it is not used anywhere in this file...

mmaelicke (changes requested):
@redhog, looks good to me, nice extension.
I think with two unittests we can bring the coverage back. Both could use the Variogram to instantiate a ProbabilisticMetricSpace with two different sample sizes. That should cover all the new code.
If you want me to do that, just go ahead and assign me to the PR, I think I will find some time towards the end of the week.

skgstat/MetricSpace.py
redhog: Oh, good catch. This is meant to create some value that's greater than the greatest possible distance.... Maybe just set ot to float max actually...

skgstat/Variogram.py
redhog: It is used here:

https://github.com/mmaelicke/scikit-gstat/blob/probabalistic-metric-space/skgstat/Variogram.py#L246

just so you can do `Variogram(..., samples=1234)`

redhog:
Btw, @mmaelicke I saw you added me to the citation! Thanks! I made an account with orcid so maybe you coul update the author list on zenodo there to point to my id https://orcid.org/0000-0002-8254-1163?

skgstat/MetricSpace.py
```diff
@@ -243,3 +255,107 @@ def dists(self):
...
         # return
         return self._dists
+
+class ProbabalisticMetricSpace(MetricSpace):
+    """Like MetricSpace but samples the distance pairs only returning a
+       `samples` sized subset. `samples` can either be a fraction of
+       the total number of pairs (float < 1), or an integer count.
+    """
+    
+    def __init__(self, coords, dist_metric="euclidean", max_dist=None, samples=0.5):
```
mmaelicke, line 265: @redhog: we need another parameter here: `seed=None`. This should be passed down to any stochastic method you use (np.random.choice` I guess) to set a `np.random.seed`. That would be great to ensure reproducible samples for unittests and, more importantly, scientific applications.

skgstat/MetricSpace.py
```diff
@@ -243,3 +255,107 @@ def dists(self):
...
+    def sample_count(self):
+        if isinstance(self.samples, int):
+            return self.samples
+        return int(self.samples * len(self.coords))
+        
+    @property
+    def lidx(self):
+        """The sampled indices into `self.coords` for the left sample."""
+        if self._lidx is None:
+            self._lidx = np.random.choice(len(self.coords), size=self.sample_count, replace=False)
```
mmaelicke, line 304: Then, here the `self.seed` can be passed down.


BTW. numpy is suggesting to use [Generator](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html#numpy.random.Generator.choice) over `np.random.choice` today...

skgstat/MetricSpace.py
```diff
@@ -243,3 +255,107 @@ def dists(self):
...
+        return self._lidx
+    
+    @property
+    def ridx(self):
+        """The sampled indices into `self.coords` for the right sample."""
+        if self._ridx is None:
+            self._ridx = np.random.choice(len(self.coords), size=self.sample_count, replace=False)
```
mmaelicke, line 311: same here

mmaelicke (changes requested):
@redhog, thanks for adding all the stuff, I think everything is addressed. There is one more thing that came to my mind, should be an easy fix
Best

skgstat/MetricSpace.py
redhog: Wait, so np.random.seed(42) does not apply "globally" to all numpy random operations after that call? That's what I did in the unittest. Maybe I've just misunderstood np.random a bit, and my code only works by accident?

mmaelicke: No, I think I was misconcepted here. The unittests work fine, so I guess that's fine. 
The reason why I like to have something like `ProbabilisticMetricSpace(..., seed=1234)` is that the seed appears as an argument to the user indicating that the class can somewhat be seeded. 
That is basically the main reason why I would like to have such an argument... 

What do you think?

redhog: Hm, maybe it _is_ better with a parameter after all, as np.random.choice is only called when a property is first evaluated so it's non-trivial to know what the seed actually is at that point?

```
s = ProbabilisticMetricSpace()
np.random.seed(42)
s.lidx, s.ridx

```

isn't a particularly good API ;)

mmaelicke (approved):
Thanks! Great PR, I really like it.

mmaelicke:
You can merge whenever you want. I will add the changelog and release the new version to PyPI, this afternoon

PR #88, opened 2021-04-16
Continuation Gstoos interface

mmaelicke:
This is the continuation of #79 to add some comments and minor changes to finally merge into `main`.


mmaelicke:
@MuellerSeb I get some numpy incompabilities here locally. I just ran it on a quite old docker container and I guess my numpy is fine for skgstat but too old for gstools.
Does gstools have a minimum requirement for numpy version, that I need to test for in the interface? 
Thanks

MuellerSeb:
The version restriction for numpy is the following:
```
numpy>=1.14.5,<2.0
```
See here:
- https://github.com/GeoStat-Framework/GSTools/blob/develop/pyproject.toml
- https://github.com/GeoStat-Framework/GSTools/blob/develop/setup.cfg

mmaelicke:
Yeah, sorry. Just realized I was running a python 3.5 container, which I should really drop. For all others, 3.6, 3.7, 3.8 it works fine.

MuellerSeb:
Yes, Python 3.5 had its EOL in September 2020, therefore we dropped support and cleaned up our code.

mmaelicke:
@MuellerSeb  I just reworked my Matern model to not change the rescaling with range. Now, I have difficulties to find the correct rescale. I think it should be `4.0` but the unittests keep failing.
I guess the solution is obvious for you...

Secondly, I just ran into an error while testing around: The `CovModel` complained about 
```
ValueError: nu needs to be >= 0.2, got: 0.00981316362270968
```
Are there mathematical implications to not accept a `nu < 0.2` that I am just ignoring? Aka. would it be wise to fix my fitting procedure to acknowlege the same limit, or is just a warning on calling `to_gstools` enough from your point of view. 
I am really not so much into Matern models, that why I am struggling here.

mmaelicke:
Ok, I tracked it down to the first bin (`matern(0)`): Your variogram yields `0` for the first bin and mine yields `np.nan`. 
I'll check my code...

MuellerSeb:
> @MuellerSeb I just reworked my Matern model to not change the rescaling with range. Now, I have difficulties to find the correct rescale. I think it should be `4.0` but the unittests keep failing.
> I guess the solution is obvious for you...

4 should be fine. Was wondering about the failing tests as well. Is the variogram evaluated at the correct locations? 

> Secondly, I just ran into an error while testing around: The `CovModel` complained about
> 
> ```
> ValueError: nu needs to be >= 0.2, got: 0.00981316362270968
> ```
> 
> Are there mathematical implications to not accept a `nu < 0.2` that I am just ignoring? Aka. would it be wise to fix my fitting procedure to acknowlege the same limit, or is just a warning on calling `to_gstools` enough from your point of view.
> I am really not so much into Matern models, that why I am struggling here.

I choose this, because the model is converging to a nugget model for `nu` close to 0. The spectral density is degenerating for very small values of `nu` and therefore I choose this lower bound. A value of `0.00981316362270968` is really just a nugget model. If you are fitting a nugget as well you will get troubles here finding a unique set of parameters.

But still I think it is not optimal that this ValueError comes up here.

mmaelicke:
@MuellerSeb, I now fixed the Matern model to return the nugget at lag 0. I encountered a similar problem in my stable model. That one was raising a `ZeroDivisionError` for too small `s`. Nice, both is fixed now.

Concerning the `nu` parameter: I just plotted a couple of models with really small smoothness, I and can see what you mean. Thus, thanks for the input and explanation and I will introduce a limit for my models as well.

MuellerSeb:
You could think about a switch that controls if the nugget is returned for `lag=0` or 0. Actually the variogram at `lag=0` *should* be always 0 but returning the nugget enables kriging routines to interpolate more smoothly when there is a nugget included.
See the discussion here for example: https://github.com/GeoStat-Framework/PyKrige/pull/153

mmaelicke:
Just released everything as `0.4.4`. I did also update the tutorial: https://mmaelicke.github.io/scikit-gstat/tutorials/06_gstools.html

PR #94, opened 2021-04-23
Gstools kriging

### LINKED ISSUE #90: Export coordinates as cond_pos
I wrapped my whole day around the skgstat - gstools interface today. I tried to use the `gstools.krige.Krige` with one of `skgstat.Variogram`.  This is what happened:
image
 
The sciki-gstat Ordinary Kriging worked just well.  After hours I finally found the source of error, which was as usually sitting in front of the computer. This is the code I used:
```Python
model = panvar.to_gstools()    # panvar is a Variogram
cond_pos = panvar.coordinates    #  <-- this is the mistake
cond_vals = panvar.values
```

The `Variogram.coordinates` returns a numpy array of shape `(n_obs, dim)`, while `Krige` wants a list of length `n_obs` for each `dim`.
Finally, the solution is simply:

```Python
x, y = zip(*panvar.coordinates)
gs.krige.Ordinary(model, [x, y], panvar.values)
```

I feel like this will not be the last time that I do this wrong. Possibly, other users of scikit-gstat might run into the same trap. I will add a note into the docs, but I think I have to be more proactive here.
My suggestion is to add a new function to `Variogram`, similar to `get_empirical` to return `cond_pos` and `cond_vals`. @MuellerSeb what do you think? Any other thought, suggestion or pathway, before I get started?

mmaelicke:
As the export from skgstat coordinates into `gstools.krige.Kriging` can easily lead to some mess (see #90), this PR includes another interface to gstools. This time, exporting a Variogram directly into a `gstools.krige.Krige` instance.
Closes #90

example:

```Python
import skgstat as skg
import pandas as pd

df = pd.read_csv('path/to/skg-docs/data/sample_sr.csv')

V = skg.Variogram(df[['x', 'y']].values, df.z.values, model='exponential', maxlag=100)
krige = V.to_krige()

x = y = range(100)
field, sigma = krige.structured((x, y))
```
image

@MuellerSeb would you mind reviewing the interface?
I copied the description for the `**kwargs` passed to Krige over from the gstools docs. Are you fine with that?


skgstat/Variogram.py
```diff
@@ -2111,6 +2111,103 @@ def to_gstools(self, **kwargs):
...
+
+        This can only export isotropic models.
+        Note: the `fit_variogram` is always set to `False`
+
+        Parameters
+        ----------
+        variogram : skgstat.Variogram
+            Scikit-GStat Variogram instamce
+        **kwargs
+            Keyword arguments forwarded to GSTools Krige.
```
MuellerSeb, line 2126: This could be a good point for a reference to the GSTools documentation with intersphinx:
https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

So you could refrain from copying the whole doc from GSTools (which might could change, as you have already pointed out)

skgstat/Variogram.py
```diff
@@ -2111,6 +2111,103 @@ def to_gstools(self, **kwargs):
         """
         return skgstat_to_gstools(self, **kwargs)
 
+    def to_krige(self, **kwargs):
```
MuellerSeb, line 2114: Maybe a more descriptive name? `to_gs_krige`? `to_gstools_krige`? But I am fine with this also. Just to make it clear for the user, that this is an external kriging routine here.

skgstat/interfaces/gstools.py
```diff
@@ -97,3 +97,125 @@ def skgstat_to_gstools(variogram, **kwargs):
...
+
+    # at least gstools>=1.3.0 is needed
+    if list(map(int, gs.__version__.split(".")[:2])) < [1, 3]:
+        raise ValueError("to_gstools: GSTools v1.3 or greater requiered.")
+
+    # convert variogram to a CovModel
+    model = skgstat_to_gstools(variogram=variogram)
+
+    # extract cond_pos and cond_vals
+    cond_pos = list(zip(*variogram.coordinates))
```
MuellerSeb, line 211: when `variogram.coordinates` is a numpy array, I would argue for `variogram.coordinates.T` since this is faster.

skgstat/interfaces/gstools.py
```diff
@@ -97,3 +97,125 @@ def skgstat_to_gstools(variogram, **kwargs):
...
+    cond_vals = variogram.values
+
+    # disable the re-fitting of the variogram in gstools
+    kwargs['fit_variogram'] = False
```
MuellerSeb, line 215: Good catch :wink:

MuellerSeb (changes requested):
I like the idea. But I would still provide an example that shows, that only the `variogram.coordinates` array needs to be transposed.

skgstat/interfaces/gstools.py
mmaelicke: agree

skgstat/Variogram.py
mmaelicke: Agree. Makes sense. I'll go for `to_gs_krige`

skgstat/Variogram.py
mmaelicke: Yeah, I'll give that a try as well.

mmaelicke:
> I like the idea. But I would still provide an example that shows, that only the `variogram.coordinates` array needs to be transposed.

Thanks for the quick review! Perfect. Then I will work on your suggestions, add another chapter to the docs, and underline the transposing issue for the `to_gstools` as well, to make that clear. In the docstring and the docs.

skgstat/interfaces/gstools.py
```diff
@@ -97,3 +107,67 @@ def skgstat_to_gstools(variogram, **kwargs):
...
+        When given Variogram model is not supported ('harmonize').
+
+    Returns
+    -------
+    :any:`Krige <gstools.krige.Krige>`
+        Instantiated GSTools Krige class.
+
+    See Also
+    --------
+    gstools.krige.Krige
```
MuellerSeb, line 146: The `Krige` class is also at top level: `gstools.Krige`.

skgstat/Variogram.py
```diff
@@ -2108,9 +2108,58 @@ def to_gstools(self, **kwargs):
...
+            When given Variogram model is not supported ('harmonize').
+
+        Returns
+        -------
+        :any:`Krige`
+            Instantiated GSTools Krige class.
+
+        See Also
+        --------
+        gstools.krige.Krige
```
MuellerSeb, line 2158: The `Krige` class is also at top level: `gstools.Krige`.

docs/tutorials/06_gstools.ipynb
```diff
@@ -13,10 +13,11 @@
    "source": [
     "With version `0.5` `scikit-gstat` offers an interface to the awesome [gstools](https://github.com/GeoStat-Framework/GSTools) library. This way, you can use a `Variogram` estimated with `scikit-gstat` in [gstools](https://github.com/GeoStat-Framework/GSTools) to perform random field generation, kriging and much, much more.\n",
     "\n",
-    "For a `Variogram` instance, there are two possibilities to export into [gstools](https://github.com/GeoStat-Framework/GSTools): \n",
+    "For a `Variogram` instance, there three two possibilities to export into [gstools](https://github.com/GeoStat-Framework/GSTools): \n",
```
MuellerSeb, line 16: Typo.

MuellerSeb (approved):
Looks good to me! :wink: Guess this is a nice shortcut! Hope the intersphinx stuff is working.

mmaelicke:
> Looks good to me!  Guess this is a nice shortcut! Hope the intersphinx stuff is working.

Thank for approval and your time!
Yeah intersphinx worked lokally, for online I will check later, then.
Thank you!

MuellerSeb:
One thing that came to my mind: is this correctly working in 1d?

PR #97, opened 2021-05-03
Update the docs to use the same cite block as README

redhog:


mmaelicke:
Thank you for pointing out

Issue #111, opened 2021-07-05
Easier support for raster data: custom bins, user-defined maxlag, return count, etc... 


rhugonnet:
Hi there!

First, thanks for all the work on scikit-gstat :+1:.
It's great and I'm very happy to see geostatistics emerging in Python!
I used `skgstat` quite a bit recently (moving away from R), cited it in our last paper! :wink: 

I work a lot with satellite data, and quantifying spatial correlations is crucial there too, especially for uncertainty estimation. 
However, due to the large number of samples, we can't calculate pairwise differences for ~ten millions points in an image. We have to subsample intelligently, and sometimes combine variograms from different grids.

For this, there is still some shackles in `skgstat` which make things a bit difficult. I intended to submit PRs with some proposed changes, but I think it's better if I first present those issues and we discuss on if and what changes would be best!

1. Choice of `maxlag`

If I want to subsample different grids, or several times a grid that has too many samples, it is currently not convenient to aggregate results over a similar bin distribution because the `maxlag` (and therefore all the bins) will be defined by the maximum distance forced by `skgstat` (note: we need to pass a `maxlag` larger than the grid size to ensure we use all samples).
Example below:

```python
import numpy as np
import skgstat

# Big grid of random data
vals = np.random.normal(0,1,size=(1000, 1000))

# Coordinates
x = np.arange(0,1000)
y = np.arange(0,1000)
xx, yy = np.meshgrid(x,y)

# Flatten everything because we don't care about the 2D at this point
coords = np.dstack((xx.flatten(), yy.flatten())).squeeze()
vals = vals.flatten()

# Run a variogram

# With a million samples, we can wait until next year...
# V = skgstat.Variogram(coordinates=(xx,yy), values=vals, normalize=False)

# Let's subsample instead, but we might need to sample from different subgrids, or we might want to run the sampling
# several times to get more robust results
nrun = 10

list_bins = []
for i in range(nrun):
    # Let's keep it to a thousand samples for speed
    subset = np.random.choice(len(vals), 2000, replace=False)
    coords_sub = coords[subset, :]
    vals_sub = vals[subset]

    # Let's use a maxlag of 2000 which is larger than the maximum grid distance
    V = skgstat.Variogram(coordinates=coords_sub, values=vals_sub, normalize=False, maxlag=2000)

    list_bins.append(V.bins)

maxlags = [list_bins[i][-1] for i in range(nrun)]
print(maxlags)
```
[...]

PR #112, opened 2021-07-06
Add custom bins (callable, iterable) and count property

rhugonnet:
Discussion that started in #111 

## Customs bins

- the `bin_func` argument has been extended to allow for `Callable` (custom function for binning) or `Iterable` (custom bin edges) as input. Default behaviour is that, when an Iterable is used, the `nlags` and `maxlag` arguments are ignored if provided by the user, and later set according to the Iterable length and maximum.
- added `test_binning_callable_arg` and `test_binning_iterable_arg` to check everything worked properly.

## Count property

- a `count` attribute is now accessible by users with the sample count of pairwise distances per bin, and automatically updated when re-computing bins with a new `bin_func`, `nlags` or `maxlag`.
- added `test_get_count` to check the above behaviour.

## To possibly add?

- check if the Callable indeed takes 3 arguments of the right shape (distances, nlag, maxlag), and returns two of the right type as well (array, float), otherwise raise error specifying the desired Callable format.
- check whether the Iterable/Callable `bin_func` returns bins in ascending lag order, and raise an error otherwise?

## PS: tests

I have 3 tests failing with `unittest` in `test_variogram.py` directly after cloning the `master` branch, not sure why!

See below
```
FAIL: test_dense_maxlag_inf (test_variogram.TestSpatiallyCorrelatedData)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/atom/code/devel/libs/scikit-gstat/skgstat/tests/test_variogram.py", line 48, in test_dense_maxlag_inf
    self.assertAlmostEqual(x, y, places=3)
AssertionError: 26.058669854641856 != 49999.99999500066 within 3 places (49973.94132514602 difference)

======================================================================
FAIL: test_sparse_maxlag_30 (test_variogram.TestSpatiallyCorrelatedData)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/atom/code/devel/libs/scikit-gstat/skgstat/tests/test_variogram.py", line 60, in test_sparse_maxlag_30
    self.assertAlmostEqual(x, y, places=3)
AssertionError: 17.129667030800196 != 17.128 within 3 places (0.0016670308001955902 difference)

======================================================================
FAIL: test_binning_iterable_arg (test_variogram.TestVariogramArguments)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/atom/code/devel/libs/scikit-gstat/skgstat/tests/test_variogram.py", line 288, in test_binning_iterable_arg
    assert V.maxlag is max(custom_bins)
AssertionError
```

### DIFF
--- a/skgstat/Variogram.py
+++ b/skgstat/Variogram.py
@@ -3,7 +3,7 @@ Variogram class
 """
 import copy
 import warnings
-from typing import Iterable, Callable, Union
+from typing import Iterable, Callable, Union, Tuple
 
 import numpy as np
 from pandas import DataFrame
@@ -525,7 +525,7 @@ class Variogram(object):
     def bin_func(self, bin_func):
         self.set_bin_func(bin_func=bin_func)
 
-    def set_bin_func(self, bin_func: Union[str, Iterable, Callable[[np.ndarray,float,float],np.ndarray]]):
+    def set_bin_func(self, bin_func: Union[str, Iterable, Callable[[np.ndarray,float,float], Tuple[np.ndarray, float]]]):
         r"""Set binning function
 
         Sets a new binning function to be used. The new binning method is set
--- a/skgstat/Variogram.py
+++ b/skgstat/Variogram.py
@@ -705,7 +705,7 @@ class Variogram(object):
         # if the input is an iterable with bin edges, we need to write self._bins here
         else:
             self._bins = bin_func
-            self._maxlag = None
+            self._maxlag = max(bin_func)
             self._n_lags = sum(1 for e in bin_func)
 
         self.cof, self.cov = None, None
--- a/skgstat/Variogram.py
+++ b/skgstat/Variogram.py
@@ -312,7 +312,7 @@ class Variogram(object):
         self._bin_func = None
         self._groups = None
         self._bins = None
-        self._count = None
+        self._bin_count = None
         self.set_bin_func(bin_func=bin_func)
 
         # Needed for harmonized models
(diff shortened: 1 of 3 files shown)


mmaelicke:
Hey man,
Thank you so much for this PR! I am rushing from meeting to meeting today, but I will try to review the PR this evening, or tomorrow the latest!
Thanks again.

skgstat/Variogram.py
```diff
@@ -801,13 +825,21 @@ def n_lags(self, n):
...
+        # reset the groups and count
         self._groups = None
+        self._count = None
 
         # reset the fitting
         self.cof = None
         self.cov = None
 
+    @property
+    def count(self):
```
mmaelicke, line 837: I like your approach very much, but I am just wondering if `bin_count` would make a better property name, to make things more clear. What do you think?

The 'internal' `_count` is fine, but could also go with `_bin_count` to make the relationship more clear

mmaelicke (approved):
Really nice, thank you again!
Even with uinttests added, that's awesome. I have just one little comment, that you can think about

skgstat/Variogram.py
rhugonnet: Changed to `bin_count` in 0e80abf
All tests passing!

mmaelicke:
Great, @rhugonnet thanks!
I'll merge this and release the new version later this day, when I am on my other computer again.
Concerning the remaining suggestions in #111 , we can switch back to that issue for further discussion. @redhog also posted an answer there.
Thanks for your contribution!

rhugonnet:
Thanks a lot for the quick feedback and release @mmaelicke! It's a pleasure to contribute a tiny bit.
For the version release, I pre-wrote a note without specifying the release number here (not sure if that was relevant): https://github.com/mmaelicke/scikit-gstat/pull/112/files#diff-ce6897be63795379cf095a62b71b40533454648a88653938c77cbd289e395547R557

I also added a very small PR to clarify optional model fitting which, actually, was already behaving close to expected when passing `fit_method = None`!

mmaelicke:
just released as `0.6.5`

PR #113, opened 2021-07-07
Argument for optional model fitting

rhugonnet:
Mentioned in #111 

## Idea
Allow an argument for the `Variogram` class to compute only the experimental/empirical variogram, and not run the fitting procedure, potentially saving some processing time.

## Current behaviour
Setting `fit_method` to `None` already skipped the fit, but performed the `preprocessing` nonetheless. This was not mentioned in the documentation.

## Changes

- Specified the Optional type for the `fit_method` argument (`None` or `str`)
- Added a sentence in the Variogram class description of the `fit_method` argument to specify that if None is passed, the fit does not run.
- `preprocessing` is skipped on Variogram initialization if `fit_func` is None; the `fit` function with `fit_method` set to None already had this behaviour
- Added a `test_nofit` to check the behaviour is correct. All tests passing


mmaelicke (approved):
Nice, thanks! I was not aware, that passing `None` already skipped the fitting. This was a coincidence and not intended behavior. 
Now it's tested and documented, thanks!

mmaelicke:
Just released as `0.6.6`

Thanks, again.

PR #128, opened 2022-09-09
fix: genton semivariance estimator bug

KentaroAoyama:
Hi guys!
A bug in the Genton semivariance estimator has been fixed in this PR.
I hope you will check it out.


mmaelicke:
Yeah, seems like you're right. I will also change the now failing unittests.

PR #135, opened 2022-12-02
Cross variogram

### LINKED ISSUE #41: Support for Cross Variogram
Support for cross variogram would be great.

mmaelicke:
This is the first PR, that actually implements a cross-variogram. With this PR, `Variogram(values=arr)`  the `arr` can be a matrix of shape `(n_samples, n_variables)`. The current state of the implementation only allows for `(n_samples, 2)`, to calculate the cross-variogram of `arr[:,0] ~ arr[:,1]`.
Upcoming versions will add a utility function to calculate cross-variograms for all combinations of multiple variables at once, and plotting thereof. 
The current changes apply to `Variogram` and `DirectionalVariogram`, **not** to `SpaceTimeVariogram`.

<hr/>
@GatorGlaciology, I used the samples of your tutorial Nr. 9: https://github.com/GatorGlaciology/GStatSim/blob/main/demos/9_cokriging_and_cosimulation_MM1.ipynb

I also co-located the surface measurements and applied the transformation, and ended up with this cross-variogram of `Nbed ~ Surface`:

image

I build the `coords` and `vals` variables in the chunk above inside your demo notebook like:
image

@GatorGlaciology, does this make sense to you? 

For reference, I put the primary variogram for `Nbed` only here as well:
image

<hr />

Also mentioning @mansourdiene10 to let you know about the current state of cross-variograms.
@digital-idiot, it's been a while, but this might also be interesting for you and should close #41

### DIFF
--- a/skgstat/DirectionalVariogram.py
+++ b/skgstat/DirectionalVariogram.py
@@ -94,11 +94,11 @@ class DirectionalVariogram(Variogram):
         bin_func : str
             .. versionchanged:: 0.3.8
                 added 'fd', 'sturges', 'scott', 'sqrt', 'doane'
-            
+
             String identifying the binning function used to find lag class
             edges. All methods calculate bin edges on the interval [0, maxlag[.
             Possible values are:
-            
+
                 * `'even'` (default) finds `n_lags` same width bins
                 * `'uniform'` forms `n_lags` bins of same data count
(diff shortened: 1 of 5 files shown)


GatorGlaciology:
This looks great! The results make sense. Thank you for creating this. We will start diving more deeply into this application

PR #155, opened 2023-07-05
GStatSim interface

mmaelicke:
@GatorGlaciology, this might be interesting for you.
This pull request introduces an preliminary interface to GStatSim. There are 3 main options for interfacing:

* generate a GStatSim compatible `prediction_grid` from a `Variogram` instance
* generate the input to a `skrige_sgs` or `okrige_sgs` simulation from a `Variogram`
* directly run `skrige_sgs` or `okrige_sgs` in parallel, from a `Variogram` instance

additionally, the interface defines a `Grid` class, which generates the `prediction_grid` but keeps track of the shape needed for the target field, to transform the result. @GatorGlaciology, maybe this class is also helpful for you? I could open a pull request to add it to GStatSim instead of SciKit-GStat, if you are interested.

#### prediction grid
You can generate a `prediction_grid` like:

```python
coords, vals = skg.data.pancake(N=300).get('sample')
vario = skg.Variogram(coords, vals, maxlag=0.6, n_lags=25)

# use like
grid = vario.gstatsim_prediction_grid(resolution=5)

# this will find rows, cols as close to 400x400 as possible to have the same resolution in both directions
grid = pred_grid = vario.gstatsim_prediction_grid(rows=400, cols=400)  
print(grid).   # prints <Grid with 401 rows and 405 cols at 1.235 resolution>

# then convert to GStatSim:
grid.prediction_grid
```

#### simulation parameters
The simulation params can be accessed via the interface

```python
from skgstat.interfaces import gstatsim_mod

# you can pass a grid, or a resolution or a (rows, cols) tuple as grid
grid, cond_df, variogram_params = gstatsim_mod.simulation_params(vario, 5)
```

#### all in one
And finally use the simulation directly

```python
fields = vario.simulate(grid, method='simple', n_jobs=8, size=8)
```

This creates a list of 8 simulated fields. If 8 cores are available, this is as fast as using GStatSim only (or n_jobs=1) as each simulation is run in parallel.

### ToDos

@GatorGlaciology, I will add a bit of docs and tests. Meanwhile you are more than welcome to comment on this interface.
I also need to define, which Variogram instances actually can be exported to GStatSim.


GatorGlaciology:
Very cool! This looks incredibly useful. Let us know if we can help in any way

PR #158, opened 2023-08-01
Fix Dowd's estimator scale

### LINKED ISSUE #157: Scale of Dowd's estimator
Hey @mmaelicke,

Hope all is well, it's been a while!
I use a lot robust estimators through my work, and noticed that Dowd's is not scaled the same as others: it returns 2gamma instead of gamma, and so will have twice the values of e.g. Matheron's for the same data. We can see it in the doc: 
https://scikit-gstat.readthedocs.io/en/latest/_images/compare_estimators.png

Easy fix dividing by 2 here: https://github.com/mmaelicke/scikit-gstat/blob/81463bbadec56e892135122856613b57ed25bea7/skgstat/estimators.py#L178
I'll open a PR!

rhugonnet:
Resolves #157 

I think that's it, didn't find anything to update in the documentation, the formulas are correct :smile:

### DIFF
--- a/skgstat/estimators.py
+++ b/skgstat/estimators.py
@@ -175,7 +175,7 @@ def dowd(x):
 
     """
     # convert
-    return 2.198 * np.nanmedian(x)**2
+    return 2.198 * np.nanmedian(x)**2 / 2
 
 
 @jit(forceobj=True)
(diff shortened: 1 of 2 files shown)


mmaelicke:
Thanks for pointing out, thanks for the fix!

PR #162, opened 2023-08-22
Skip GSTools test in `test_variogram` if not installed

rhugonnet:
I noticed this failed during my tests in #160! Also saw the comment here: https://github.com/mmaelicke/scikit-gstat/pull/148#issuecomment-1625059544

This should fix it, logic copied from `test_interfaces` :smile:


rhugonnet:
Fixed minor conflicts with the new main, ready to merge!

PR #163, opened 2023-08-22
Check code and docs spelling via `pre-commit`

rhugonnet:
Hey @mmaelicke,

I've been noticing typos here and there in the package. In other projects, I also had a hard time with those for a while until I started using `pre-commit`/`codespell` (and other hooks).

I ran it out of curiosity on `skgstat`, and it's a pretty long output! Opening this temporary PR for you to see the output of it in CI: https://github.com/mmaelicke/scikit-gstat/actions/runs/5945047993/job/16123425002?pr=163

There's also a ton of other useful things that can be done in `pre-commit`, I don't how familiar you are with it. I had trouble starting a couple years ago, and now I can't code (in a repo) without, it saves so much time down the line!

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,15 @@
+name: Linting and formatting (pre-commit)
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+    - uses: pre-commit/action@v3.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+        # Fix common spelling mistakes
+        - repo: https://github.com/codespell-project/codespell
+          rev: v2.2.1
+          hooks:
+                - id: codespell
+                  args: [
+#                    '--ignore-words-list', 'nd,alos,inout',
+#                    '--ignore-regex', '\bhist\b',
+                    '--'
+                  ]
+                  types_or: [python, rst, markdown]
+                  files: ^(skgstat|docs|tutorials)/


mmaelicke:
HI @rhugonnet, this is great! I was not aware of these hooks, thanks for pointing me at this. 

This is really neat at also necessary for SciKit-GStat. How should we go on from here? Fix all the typos on this branch, and then add the pre-commit action to all CI, to only accept PRs that have been checked?

rhugonnet:
Glad you like it! Yes exactly what you said to go from here.

For contributors that want to fix their PR locally, they simply need it installed: `pip install pre-commit`. And then they can run it manually from the folder `scikit-gstat/` using `pre-commit run --all` (which calls `.pre-commit-config.yaml`) and then commit + push the changes. Or, for "advanced" users, they can configure it to be a Git hook (triggers before comitting).
The procedure for contributors could be added to a `CONTRIBUTING.rst` at the root of the package.

Do you want to do the fixes? I don't mind, but I'm already used to `pre-commit`. Maybe you want to try it out for yourself :wink:

rhugonnet:
And for info, there's quite a lot of other useful hooks, quick summary below. Most of these hooks perform changes automatically in the code because there is no room for error in their functionality (unlike `codespell` that only proposes the correct spelling to the user, because it might not know a term!). In that case, running `pre-commit` simply slightly reformats things automatically.

**Gentle hooks (small changes)**
- The common, out-of-the-box `pre-commit-hooks`: https://github.com/pre-commit/pre-commit-hooks. I typically use  **check-yaml**, **end-of-file-fixer**, **trailing-whitespace**, and **check-merge-conflict**, which all enforce basic checks.
- The hook https://github.com/MarcoGorelli/absolufy-imports, which automatically removes relative import syntax that is discouraged, e.g. "from . import models", in favor of "from skgstat import models",

**Classic hooks (moderate changes)**
- The https://github.com/pycqa/flake8 hook enforces the code to respect PEP standards, which can prevent some errors down the line.
- The https://github.com/PyCQA/isort hook sorts the imports by category at the top of files (Python, external, internal) to easily identify which are which.
- The https://github.com/asottile/pyupgrade will automatically upgrade the Python syntax to a certain version (useful when new Python versions come out!).

**Aggressive hooks (big changes, but really worth it in the long run!)**
- The https://github.com/psf/black hook will reformat all the code following strict Python standards and a couple user arguments (e.g., maximum line length). It's really powerful but "uncompromising", see their GitHub page.
- **(Most powerful hook but quite a bit of work)**: The https://github.com/python/mypy hook will enforce static type checking for the entire package, which makes it really robust to type/shape/logical errors. Since a couple years, NumPy array shapes and dtypes are supported (https://numpy.org/devdocs/reference/typing.html#mypy-plugin), which means that the shape/dtype casting of all array operations in the package can be checked statically, before running any code.

rhugonnet:
How should we move forward on this @mmaelicke? We can add only `codespell` for now and do the typo corrections, and refer the above comment with other hooks in an issue for "improvement"?

It seems like pre-commit is really becoming the norm everywhere for static checking, would make `scikit-gstat` even more robust to errors :wink:. Also, packages like pandas are using it to do a lot of other things, like having consistent environment files between `conda` and `pip` by calling a custom script: https://github.com/pandas-dev/pandas/blob/main/.pre-commit-config.yaml#L267. I recently added this in the repos I maintain, and it's an amazing help to keep environment files consistent everywhere!

mmaelicke:
Yeah, sorry for the delay. I was busy last week and not close to a computer and have some stuff on the agenda for today and tomorrow. I can implement the `pre-commit` stuff towards the end of the week along with solving the numba issues.

PR #176, opened 2024-01-18
adding RSS and RMS

sebasmos:
Contributions: 

included and tested the metrics RSS and RSM, it is suggested in the reference below

Reference: 

Oliver, M. A., & Webster, R. (2015). Basic steps in geostatistics: the variogram and kriging (No. 11599). Cham, Switzerland: Springer International Publishing.

Link: https://link.springer.com/book/10.1007/978-3-319-15865-5

### DIFF
--- a/skgstat/Variogram.py
+++ b/skgstat/Variogram.py
@@ -2219,6 +2219,7 @@ class Variogram(object):
             DeprecationWarning
         )
         return self.model_residuals
+    
 
     @property
     def model_residuals(self) -> np.ndarray:
@@ -2382,6 +2383,33 @@ class Variogram(object):
 
         """
         return self.rmse / np.nanmean(self.experimental)
+    @property
+    def root_mean_square(self):
+        """Root Mean Square (RMS) of the residuals
+
+        Calculates the square root of the mean of squared residuals.
+
+        Returns
+        -------
+        float
+            Root Mean Square of the residuals.
+        """
+        return np.sqrt(np.nanmean(np.square(self.model_residuals)))
+    @property
+    def residual_sum_of_squares(self):
+        """Residual Sum of Squares (RSS)
+
+        Calculates the sum of squared differences between the experimental
+        variogram and theoretical model values.
+
+        Returns
+        -------
+        float
+            Residual sum of squares (RSS), a measure of the overall model fit
+            representing the sum of squared deviations between the observed
+            experimental variogram and the corresponding theoretical model values.
+        """
+        return np.nansum(np.square(self.model_residuals))
 
     @property
     def nrmse_r(self):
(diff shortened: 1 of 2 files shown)


skgstat/Variogram.py
```diff
@@ -2383,6 +2384,39 @@ def nrmse(self):
...
+        """Root Mean Square (RMS) of the residuals
+
+        Calculates the square root of the mean of squared residuals.
+
+        Returns
+        -------
+        float
+            Root Mean Square of the residuals.
+        """
+        return np.sqrt(np.nanmean(np.square(self.model_residuals)))
```
mmaelicke, line 2398: Sorry if I'm blind right now, but is this not the same as `Variogram.rmse`?
We could update the the existing rmse with this implementation, which is way cleaner; or let `Variogram.rmse` just call your function.

skgstat/Variogram.py
```diff
@@ -2383,6 +2384,39 @@ def nrmse(self):
...
+        float
+            Residual sum of squares (RSS), a measure of the overall model fit
+            representing the sum of squared deviations between the observed
+            experimental variogram and the corresponding theoretical model values.
+        """
+        return np.nansum(np.square(self.model_residuals))
+
+    @property
+    def rss(self):
+        return self.residual_sum_of_squares
```
mmaelicke, line 2418: I added this, cause I like to have both, the full name and the short name.

mmaelicke (changes requested):
Hi,
Thanks a lot for your contribution!
I have a minor comment that needs clarification.

skgstat/Variogram.py
sebasmos: It seems like mean square of the residuals (RMS) could be the Root Mean Square Error (RMSE), however in the book [1](https://link.springer.com/book/10.1007/978-3-319-15865-5) its not explicitaly menctioned ~ which is why I added it separately; similarly here is used with the same nomenclature [2](https://www.researchgate.net/figure/Estimation-of-the-root-mean-square-RMS-of-the-residual-between-the-data-and-the-model_fig5_326735330). 

It could be a typo in the book (?) and its just the RMSE, for me leaving RMSE is fine..

skgstat/Variogram.py
sebasmos: Sounds great

skgstat/Variogram.py
mmaelicke: I pointed my old `Variogram.rmse` to your `Variogram.root_mean_square` and all the tests still work, so in fact, it is the same thing.
Then, we keep both with `Variogram.rmse` simply calling `Variogram.root_mean_square`, as it has the much cleaner implementation.

mmaelicke:
Thanks again. Just published as `1.0.16` on PyPI.

PR #189, opened 2024-08-30
Bug/wrapped distance function variogram

frank-hulo:
Found small bug after my last update

Test are running as well


skgstat/Variogram.py
```diff
@@ -1149,10 +1149,7 @@ def dist_function(self):
 
     @classmethod
     def wrapped_distance_function(cls, dist_func, x, **kwargs):
-        if callable(dist_func):
-            return dist_func(x, **kwargs)
-        else:
-            return pdist(X=x, metric=dist_func, **kwargs)
```
mmaelicke, line 1155: Hi man. I'm on vacation without my Laptop, using just my phone, sorry. Could you quickly guide me through these changes? Why is the check for callables not necessary anymore?

frank-hulo: Hi, I hope you're enjoying your vacation! The behaviour currently isn't the same as what you're describing. From what I understand, the method you're referring to seems to be limited to Kriging.py. In MetricSpace, the metric is passed with kwargs into `pdist`, while in the variogram, the metric is called directly if it's a callable. In my opinion, if the metric is callable, it should behave the same way as it does in MetricSpace. `pdist` can handle the metric if it is a string or callable.

frank-hulo: This is how dists is implemented in MetricSpace:
```
    @property
    def dists(self):
        """A distance matrix of all point pairs. If `self.max_dist` is
        not `None` and `self.dist_metric` is set to `euclidean`, a
        `scipy.sparse.csr_matrix` sparse matrix is returned.
        """
        # calculate if not cached
        if self._dists is None:
            # check if max dist is given
            if self.max_dist is not None and self.dist_metric == "euclidean":
                self._dists = self.tree.sparse_distance_matrix(
                    self.tree, self.max_dist, output_type="coo_matrix"
                ).tocsr()

            # otherwise use pdist
            else:
                self._dists = squareform(
                    pdist(self.coords, metric=self.dist_metric, **self.dist_metric_kwargs)
                )

        # return
        return self._dists
```

mmaelicke:
published as `v1.0.19`

PR #192, opened 2025-01-27
Update docstring spherical model

pepijnvtol:
Add `\left` and `\rigth` to docstring of spherical model. On the docs it looks like that `\frac{h}[a}^3` says that only `h^3`, but the source is correct and it should be `(h/a)^3`.

$$
\frac{h}{a}^3 \text{vs } \left(\frac{h}{a}\right)^3
$$

### DIFF
--- a/skgstat/models.py
+++ b/skgstat/models.py
@@ -54,7 +54,7 @@ def spherical(h, r, c0, b=0.0):
     The implementation follows [6]_:
 
     .. math::
-        \gamma = b + C_0 * \left({1.5*\frac{h}{a} - 0.5*(\frac{h}{a})^3}\right)
+        \gamma = b + C_0 * \left({1.5*\frac{h}{a} - 0.5*\left(\frac{h}{a}\right)^3}\right)
 
     if :math:`h < r`, and
 
--- a/skgstat/models.py
+++ b/skgstat/models.py
@@ -98,7 +98,7 @@ def exponential(h, r, c0, b=0.0):
     r : float
         The effective range. Note this is not the range parameter! For the
         exponential variogram function the range parameter a is defined to be
-        :math:`a=\frac{r}{2}`. The effective range is the lag where 95% of the
+        :math:`a=\frac{r}{3}`. The effective range is the lag where 95% of the
         sill are exceeded. This is needed as the sill is only approached
         asymptotically by an exponential function.
     c0 : float
@@ -160,7 +160,7 @@ def gaussian(h, r, c0, b=0.0):
     r : float
         The effective range. Note this is not the range parameter! For the
         exponential variogram function the range parameter a is defined to be
-        :math:`a=\frac{r}{3}`. The effetive range is the lag where 95% of the
+        :math:`a=\frac{r}{2}`. The effetive range is the lag where 95% of the
         sill are exceeded. This is needed as the sill is only approached
         asymptotically by an exponential function.
     c0 : float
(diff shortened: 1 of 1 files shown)


pepijnvtol:
For the Gaussian model:

When the following equation is used:
$$\gamma(h) = b + C_0 \cdot \left(1 - e^{-\frac{h^2}{a^2}}\right),$$

and for the range, it is stated that:
_The effective range is the lag where 95% of the sill is exceeded,_ 

we want to solve for $\left(1 - e^{-\frac{h^2}{a^2}}\right) = 0.95.$

This leads to:
$$0.05 = e^{-\frac{h^2}{a^2}},$$
which simplifies to:
$$\ln(20) = \frac{h^2}{a^2},$$
and finally:
$$h = \sqrt{\ln(20)} \cdot a \approx 1.73 \cdot a.$$

skgstat/models.py
```diff
@@ -54,7 +54,7 @@ def spherical(h, r, c0, b=0.0):
     The implementation follows [6]_:
 
     .. math::
-        \gamma = b + C_0 * \left({1.5*\frac{h}{a} - 0.5*(\frac{h}{a})^3}\right)
+        \gamma = b + C_0 * \left({1.5*\frac{h}{a} - 0.5*\left(\frac{h}{a}\right)^3}\right)
```
mmaelicke, line 57: Thanks for fixing the docs!

skgstat/models.py
```diff
@@ -98,7 +98,7 @@ def exponential(h, r, c0, b=0.0):
     r : float
         The effective range. Note this is not the range parameter! For the
         exponential variogram function the range parameter a is defined to be
-        :math:`a=\frac{r}{3}`. The effective range is the lag where 95% of the
+        :math:`a=\frac{r}{2}`. The effective range is the lag where 95% of the
```
mmaelicke, line 101: Long ago, that I wrote this stuff, but I think this was initially correct. the connection between effective range and range parameter should be 3a = r.

mmaelicke:
> This leads to: 0.05 = e − h 2 a 2 , which simplifies to: ln ⁡ ( 20 ) = h 2 a 2 , and finally: h = ln ⁡ ( 20 ) ⋅ a ≈ 1.73 ⋅ a .

Yeah, you are correct. 
However, I always sticked to to using a factor of 2, because it's easier to remember and implement. I also like the transition between exponential, spherical and gaussian model, to use 1a, 2a and 3a respectively. At least in my field of expertise this is common thing to do, that's why I wanted to stick to that.
While you are correct in a mathematical strict sense, my implementation leads to using something closer to 98% (I think) of the sill. That can result in a slight overestimation of the effective range of the model. However, in my applications, that did usually not make a difference. Remember, that scikit-gstat fits the effective range directly, so this parameter usually reflects the data properties.
If you need a more exact model, you could pass your own function to the Variogram class.

mmaelicke:
Yeah, sorry, you are right. It is spherical, gaussian, exponential, of course.

I am a bit confused, as that is what the docs say and your PR wants to change the r/3 for the exponential in line 101 to r/2, right? Which we both agree would not be correct. I feel like I am missing something here.

https://github.com/mmaelicke/scikit-gstat/blob/21ef9233c6df8a9d94672e659053b272662736fe/skgstat/models.py#L101

pepijnvtol:
Sorry, my mistake! You are right. I meant to change the Gaussian model, but I mistakenly modified the exponential model instead. I will update the pull request.

mmaelicke:
Thanks!

PR #195, opened 2025-05-12
Fix: Variogram - Add space to error message

jomey:
Error message needs an additional space at the end when splitting over two lines.


mmaelicke:
good catch!
sorry for the delayed response, thank you very much!

